PR #3, opened 2023-03-20
Removed some auth, added some auth, parties are strings

Teifion:


src/schema/types.ts
```diff
@@ -23,7 +23,7 @@ export const user = Type.Object({
             sync: Type.Record(Type.String(), Type.Number()),
             faction: Type.String(),
             lobby_id: Type.Integer(),
-            party_id: Type.Integer(),
+            party_id: Type.String(),
```
Jazcash, line 26: not a problem, but curious why strings for party_id?

src/schema/auth.ts
```diff
@@ -4,29 +4,12 @@ import { Endpoint } from "../helpers";
 import { privateUser } from "./types";
 
 export const authEndpoints: Record<string, Endpoint> = {
-    get_token: {
```
Jazcash, line 7: I get that you want to do auth via HTTP, but I would still like to have the schema for that defined somewhere. I assume there's also gonna be a need for other unauth'd endpoints via HTTP, e.g. account recovery, so maybe we could have a dedicated file just for those schemas?

Teifion: I'd be happy to do that. I'll add a file for it later today/tomorrow.

PR #13, opened 2023-05-03
Added command structures

Teifion:


Teifion:
As per https://github.com/beyond-all-reason/tachyon/discussions/10

Jazcash (commented):
I'm not so sure about this. Could you give some examples of lobby commands and menu commands? I think I'd rather just have these explicitly defined in the protocol.

Teifion:
Lobby commands would be things that are currently handled by either spads or the server within the lobby. For example setting min/max ratings or calling nextmap. While we want to have the GUI for them the way they'd be communicated would be via the commands system thus making it more extensible and customisable. I would suggest we define limits on the syntax for these commands as we go forwards to prevent it becoming too unweildy.

Menu commands would be things like linking your discord to the server or getting the `whoami`. We don't have many of these yet but I'm thinking about if we wanted to add more. Again, where possible we want GUI but having the ability to use slash commands etc would be good I think.

Jazcash:
That's what I was talking about the other day with discord-style `/` integration. Where all the commands are defined in the protocol as regular requests/responses, and it's up to the client app for which ones it wants to integrate into the command list, if that makes sense.

For example:
```ts
minRating: {
    request: Type.Object({
        rating: Type.Number({ minimum: 0, maximum: 50 }),
    }),
    response: Type.Object({
        result: Type.Literal("success")
    })
}
```
Then the client app can choose whether to include this command in the chat's `/` suggestions. I'd much prefer all the commands are defined in this protocol rather than being dynamically configurable on the server and runtime-evaluated, because that keeps everything in one place and keeps the command's schema definitions synchronised between server/client, and also allows type-checking at compile time.

I really think we should rethink the way autohost/server commands work for battles in general though, because right now it's a mess, essentially because we lack Perl SPADS devs to implement all the commands we want so we've hacked them into the middleware server instead. Going forward, we can do this much better. I'll open a separate discussion for that.

Teifion:
I'm still in favour of the additional flexibility having this would allow us but if you feel strongly then I'm happy to concede on this one.

Jazcash:
> I'm still in favour of the additional flexibility having this would allow us but if you feel strongly then I'm happy to concede on this one.

I'm also willing to concede if and when we have to, but I'd like to try defining everything in this protocol schema first and see how we get on.

Teifion:
I'll update the pull request later today.

PR #19, opened 2023-06-18
Various bits

Jazcash:
- Changed generation schema syntax to make the type resolution simpler and more obvious
- Responses no longer have a `data` property by default
- Removed `user` data from `getToken` endpoint. I think that should only be sent on `login`
- `id` -> `userId`. Good to be explicit with any ids in the protocol imo, as there could be multiple ids as part of the same response
- `name` -> `username`. I think `name` implies either real name or display name, but the actual usage right now is as a unique identifier, so I think `username` is more appropriate
- added `dev` script so the schema gets automatically generated on file changes (run with `npm run dev`)
- added `email` field to `PrivateUserClient` type


Teifion:
> Removed user data from getToken endpoint. I think that should only be sent on login
Does that mean ones that had an empty one no longer have it but those with it (e.g. `whoamiResponse` ) will still have one?

> Removed user data from getToken endpoint
I'm looking at the files changed but can't see an update to the documentation regarding establishing a connection. Could you alter and extend that part to ensure I correctly understand how you see this information being transmitted etc please.

> id -> userId
I 100% agree with this

> name -> username
Also a good change

> added email field to PrivateUserClient type
Thumbs up to this too

Jazcash:
> Does that mean ones that had an empty one no longer have it but those with it (e.g. `whoamiResponse` ) will still have one?

I think you misquoted me here, but yes previously all endpoints had a `data` object. I updated the `whoAmI` response to support the data object, which allows any properties in it (I think regardless of `additionalProperties: true`:

```
whoAmI: {
    request: EmptyObject;
    response: { status: "success"; data: object };
};
```

> I'm looking at the files changed but can't see an update to the documentation regarding establishing a connection. Could you alter and extend that part to ensure I correctly understand how you see this information being transmitted etc please.

Misquoted again maybe? I haven't touched anything in regards to how a connection should be established.

Also I've just removed the `permissions` property of the `PrivateUserClient` type as part of this PR also.

Teifion:
> Data

Yes I did misquote, I was rushing and that was very careless. I support the removal of empty data objects where they are not needed so that's all good.

> Misquoted again maybe? I haven't touched anything in regards to how a connection should be established.

It wasn't a quote of what you said but I put it as a quote, maybe I'm more tired than I thought. You changed things about the getToken endpoint but I can't see anything in the getToken endpoint docs that's changed to accommodate this.

I fully support the removal of the `permissions` field.

Jazcash:
Nothing in that doc should be affected by my changes, although it seems like some mismatches were already there. As discussed on Discord, we need to decide if we want to use pure Websockets or a mixture of HTTP and Websockets, but that's not relevant to this PR I think

PR #21, opened 2023-07-07
Overhaul

Jazcash:
Main things:

- Only a few endpoints for now, can build them up again as we go, but I'd rather flesh them out properly instead of creating empty stubs for all of them
- Descriptions and examples can be added in endpoint definitions and these will be used to generate docs
- Output schema now lives in `dist/`
- `order` property can be added to endpoint definitions to define the order which they appear in the docs
- Removed `clanTag` from the `battleStatus` type, as this is already part of `userClient`
- Some significant changes to the `lobby` type in `types.ts`. I'd recommending checking the diff here but notably:
    - Removed `ip`, `port` (I'd like to have autohosts send the script.txt to clients on launch instead)
    - Removed `natType` (Do clients need to know this? Happy to add it back as and when needed)
    - Removed the `settings` object (I'd like to explicitly add all the settings to the main object, or otherwise nest them, but I want to avoid untyped objects where anything goes)
    - Added a few example settings (`minTeamsize`, `maxTeamsize`, `minRating`, `maxRating`)
    - Replaced `passworded` and `locked` with `private`
    - Removed `public`
    - Removed `type` and added `replay`, happy to add `type` back if we get more than 2 types of lobby but for now I'd like to keep it simple
    - `engineName` and `engineVersion` replaced with `engine`
    - `mapName` and `mapHash` replaced with `map`
    - Added `game` to represent game version


Teifion:
For context, this is what I'm expecting the internal representation of the lobby to look like once I merge the two processes: https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/lobby/schemas/lobby_struct.ex

The exact structure is likely to change but the important attributes should be appearing there.

Democracy as a true/false seems like we'll hit issues with that pretty fast, I suspect we want something more nuanced.

What is the expected behaviour of private? We currently have locked, passworded and in future might want to go with things like "friends only" or "clan only".

startAreas look like they have to be rects, can we make it so we leave the door more easily open for other shapes in the future?

the min/max rating stuff I expect to be part of a larger object called "limits" or something like that. I fully expect there to be other limiters we want to put in place and they'll all be optional.

Jazcash:
> For context, this is what I'm expecting the internal representation of the lobby to look like once I merge the two processes: https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/lobby/schemas/lobby_struct.ex

Ah nice, I guess there's quite a bit here we'll want to make accessible to clients but we can work on that as we go.

> Democracy as a true/false seems like we'll hit issues with that pretty fast, I suspect we want something more nuanced.

Yeah maybe. Can you give some examples of issues you forsee which might help us think of a better way?

> What is the expected behaviour of private? We currently have locked, passworded and in future might want to go with things like "friends only" or "clan only".

You're right that there's nuance with different types of private games so I think I'll revert this change for now.

> startAreas look like they have to be rects, can we make it so we leave the door more easily open for other shapes in the future?

Good idea, but I'd rather do that when the time comes so we can keep the protocol as simple as it needs to be and when we know the format that other start area shapes might use (likely an array of points but would still like to wait).

> the min/max rating stuff I expect to be part of a larger object called "limits" or something like that. I fully expect there to be other limiters we want to put in place and they'll all be optional.

Sounds good. The main thing I'm getting at with those properties is having them all explicitly specified in the protocol, which means we'll likely want to add sub-blocks to the `lobby` object to help categorise them all. I'll add these to a `limits` property like you suggested.

Generally speaking, I'd like to try and simplify the idea of custom lobbies a bit in favour of prioritising and pushing matchmaking. I think trying to achieve full feature parity with SPADS would be a bad idea in terms of our time, but also perhaps a waste if matchmaking does end up being the preferred way to play like I hope it will.

Btw I added some matchmaking endpoints and also fleshed out the generated docs a lot more in my recent commits. This PR is a bit unwieldy and I don't expect you to review the entire thing and keep on top of it as I keep pushing more to it. As we're still in the early stages, it might be better to get it merged (if you approve of the more top level changes) and then raise issues/PRs for the parts you're not sure about afterwards? I'll try to keep future PRs more limited in scope but I went a bit nuts on this one because I cba to keep switching branches 😅

Teifion:
> Yeah maybe. Can you give some examples of issues you forsee which might help us think of a better way?

Some examples of different voting:
- Democracy
- Autocracy (boss mode with multiple bosses, would probably want a list of the bosses)
- Dictatorship (would be useful to know who the boss is)
- Weighted democracy (I can't see us using it but it's an example where it stops being a true/false)

Not to mention we might want to have different rules about who can vote on what for certain lobby types (e.g. tournaments).

> Good idea, but I'd rather do that when the time comes so we can keep the protocol as simple as it needs to be and when we know the format that other start area shapes might use (likely an array of points but would still like to wait).

Fair enough, just seems more sensible to have that "shape type" as it were present at this stage to prevent recoding the entire function; makes it harder to break when we refactor it later.

> Generally speaking, I'd like to try and simplify the idea of custom lobbies a bit in favour of prioritising and pushing matchmaking.

Custom lobbies are going to be a big part of the game, based on some research I've been doing I actually expect them to be bigger than matchmaking by quite a bit and matchmaking itself will be making use of them so we'll need to have support for them sorry.

I'll approve the PR on the basis I want to keep stuff moving and don't want to slow you down. Once you're happy with it and generate the new specs I can run them through my Hailstorm tests and see what does and doesn't work etc.

PR #23, opened 2024-01-24
added auth diagrams

Jazcash:


Jazcash:
btw you can display the markdown preview in the Files changed tab like so:
image

Jazcash:
Gonna merge, if you notice anything wrong with these diagrams later then PR to fix would be appreciated 🙏

PR #25, opened 2024-06-01
Add documentation on how to establish WebSocket connection

p2004a:


docs/connection.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Authorization
+
+The WebSocket connection requires authorization. The full auth process is detailed [here](authorization.md). The `Authorization` header must be set to the access token obtained from the auth process following the `Bearer` scheme.
+
+## Tachyon version negotiation
+
+The client must specify as list of supported tachyon versions in the `Sec-WebSocket-Protocol` header. The server will then select the highest version supported by both the client and server. If the server does not support any of the versions specified by the client, the connection will be closed.
+
+The name of the websocket subprotocol is `{version}.tachyon.beyondallreason.dev`.
```
Jazcash, line 23: Why does `beyondallreason.dev` need to be part of this string?

Jazcash (approved):

docs/connection.md
p2004a: Quoting WebSockets RFC, emphasis mine:

> These subprotocol names **should** be registered as per Section 11.5. To avoid potential collisions, it is **recommended to use names that contain the ASCII version of the domain name of the subprotocol's originator**. For example, if Example Corporation were to create a Chat subprotocol to be implemented by many servers around the Web, they could name it "chat.example.com".

We don't want to register them realistically, but use the domain name following recommendation.

We don't need to, it's only a recommendation, I can drop it.

Jazcash: I think that would be best

p2004a: Removed

Jazcash (approved):

PR #27, opened 2024-06-15
Upgrade to latest TS beta to fix filter typing inference

Jazcash:


Jazcash:
Idk why CI can't figure this out but it's not a big enough deal to care about imo, can just update when TS 5.5 is officially released

PR #32, opened 2024-08-31
Specify queues

geekingfrog:
Add `queueId` to requests and events

Each queue is treated independently so every event and request specific
to a given queue needs to include the ID of the queue.

### DIFF
--- a/docs/schema/matchmaking.md
+++ b/docs/schema/matchmaking.md
@@ -52,9 +52,15 @@ Cancel queueing for matchmaking.
     "properties": {
         "type": { "const": "request" },
         "messageId": { "type": "string" },
-        "commandId": { "const": "matchmaking/cancel" }
+        "commandId": { "const": "matchmaking/cancel" },
+        "data": {
+            "title": "MatchmakingCancelRequestData",
+            "type": "object",
+            "properties": { "queueId": { "type": "string" } },
+            "required": ["queueId"]
+        }
     },
-    "required": ["type", "messageId", "commandId"]
+    "required": ["type", "messageId", "commandId", "data"]
 }
 
 ```
@@ -67,7 +73,10 @@ Cancel queueing for matchmaking.
 {
     "type": "request",
     "messageId": "magna commodo",
-    "commandId": "matchmaking/cancel"
+    "commandId": "matchmaking/cancel",
+    "data": {
+        "queueId": "magna commodo"
+    }
 }
 ```
 </details>
@@ -78,6 +87,10 @@ export interface MatchmakingCancelRequest {
     type: "request";
     messageId: string;
     commandId: "matchmaking/cancel";
+    data: MatchmakingCancelRequestData;
+}
+export interface MatchmakingCancelRequestData {
+    queueId: string;
 }
 ```
 ### Response
@@ -262,8 +275,11 @@ Server should send this when players ready up using [ready](#ready).
         "data": {
             "title": "MatchmakingFoundUpdateEventData",
             "type": "object",
-            "properties": { "readyCount": { "type": "integer" } },
-            "required": ["readyCount"]
+            "properties": {
+                "queueId": { "type": "string" },
+                "readyCount": { "type": "integer" }
+            },
+            "required": ["queueId", "readyCount"]
         }
     },
     "required": ["type", "messageId", "commandId", "data"]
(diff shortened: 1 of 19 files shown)


Jazcash:
Is it wise to allow multiple queues at the same time, as separate flows? My original intention was to allow queueing for multiple playlists at once, but as part of a single flow. That is, the user would check all the playlists they wouldn't mind playing, and then they press the queue button, and then subsequent matchmaking commands would all be related to that one queue. If the user wishes to change the playlists they're queueing for, they would have to cancel the queue, check/uncheck the desired playlists and then queue again.

I'm a bit worried about the complexity of allowing multiple queues, both for the server and the client frontend, and am not too sure of the benefits.

geekingfrog:
I'm open to change that back, and then I have a few questions.

## 0) playlist
I'm confused by the term, what do you mean by that?
In my mind, a matchmaking queue is for a specific of match, for example, one queue for `1v1` and another for `2v2`.

## 1) queue requests semantics
Let's assume there are 3 available queues: `1v1`, `2v2` and `3v3`. The player select `1v1` and `2v2` and sends a `queue` request to  the server. Then the player changes their mind and select `1v1` and `3v3`. What should be the semantics if they send another `queue` request with that?
Is the intention to reject that request wholesale with a `already_queued`? That's easy on the server.
The cancel request would then also be for all the selected queues I believe?

What should happen if the player sends a `queue` request for `3v3` and `4v4` queue that doesn't exist? It seems the request would be rejected with `invalid_queue_specified`? That is a bit harder to do on the server because each queue is an independent entity with no knowledge of the other queues.

## 2) found, lost and back to queuing
Back to queuing for `1v1` and `2v2`, when a match is found, there is the id of the queue, so the player know they are getting thrown into a `2v2` for example. Does that mean that they are withdrawing from `1v1` at that time?

When the pairing fails for the `2v2` match because  of another player, I am planning to requeue the player on that same queue, with the same parameter, so they are prioritized for another pairing since they've been waiting for a while now.
Should the server does the same for the `1v1` queue? That may be somewhat difficult, although I have an idea for that.

Jazcash:
## Terminology

The terminology in my head is a _playlist_ is the possible modes a player can _queue_ for, e.g. 1v1, 2v2, ffa. A queue is the act of searching for one or more given playlists. In this instance, I'm suggesting there only be a single queue allowed that searches for multiple playlists, as opposed to multiple queues allowed where each queue only searches for 1 playlist at a time. I was using this terminology in new lobby codebase already but realised I hadn't actually defined it anywhere, if you're fine with it I'll add it to the terminology section in this repo's readme and change the schemas to match, e.g. `invalid_queue_specified` -> `invalid_playlist_specified`.

##  Queue requests semantics

Yeah exactly, I'm proposing to only allow one queue.

If any of the playlistIds is invalid in the queue request, I would expect the server to send that `invalid_playlist_specified` error, so I guess it would mean the server would need to validate the playlistIds up front before initiating the queue.

My thinking is the playlists would be defined on the server, where moderators or whoever has perms can modify them to set the playlist name, number of teams, min/max number of players per team, whether it should be ranked or not and an associated map pool.

When the user goes to the matchmaking section in the client, the client would fire off a `list` request to learn about the possible playlists it can queue for. I think a possible scenario for this `invalid_playlist_specified` occurring would be if the playlists then get changed while a user is on the matchmaking screen, and should the client receive this error, it would fire off a new `list` request to update the playlists shown.

## 2) Found, lost and back to queuing

With only one possible queue allowed, my expectation would be for the server to prompt a `found` event as soon as it finds a match for any of the playlists specified in the queue. At this point I think they should be withdrawn from the queue, to prevent any weirdness with multiple playlist matches being prompted. Should this match be invalidated (e.g. because one of the players failed to ready up), the server would send a `lost` event and restart the queue again, searching for all the playlists they specified originally. 

How the server handles that requeuing process is up to the server, but yeah I think some sort of priority system would be awesome, so people waiting the longest have a better chance of finding a match.

In my head I was thinking a queue would have a set of internal criteria, e.g preferred ping, rating, teamsize. When a player queues, the criteria might be 0-50 ping, 0-10% rating deviation. Then after x amount of time with no match, the criteria would broaden. This x amount of time before the criteria broadens could keep ticking even when a match is found, so when they get requeued they don't start back at square one. Thoughts?

geekingfrog:
Ok that makes more sense. That means we should also revert https://github.com/beyond-all-reason/tachyon/pull/31.
As for the terminology, I'm not a fan of the term playlist, I associate that more with music I guess. I think `queue` is a better name, and clients can be in different queues at the same time, but ultimately only be matched in one at a time.

Not a hill I'll die on though. The server is likely to keep the name `queue` though as quite a bit is already in place with that name, but that's only internal. If I need to change the replies to mention playlist that's trivial.

PR #34, opened 2024-09-11
Add matchmaking/cancelled event

geekingfrog:
Adding an event to boot a player out of matchmaking. 2 reasons I can see where this is needed:

1. a party member leave the matchmaking, and thus other party members are evicted
2. server's component crash. If something goes wrong and the server cannot recover the matchmaking state of the player, it's better to eject them so they can retry/reset.

### DIFF
--- a/src/schema/matchmaking/cancelled.ts
+++ b/src/schema/matchmaking/cancelled.ts
@@ -1,0 +1,19 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description:
+        "Server may send this event at any point when the user is queuing to indicate that the user has been booted out the matchmaking system.",
+    event: {
+        data: Type.Object({
+            reason: Type.Union([
+                Type.Literal("intentional"),
+                Type.Literal("server_error"),
+                Type.Literal("party_user_left"),
+            ]),
+        }),
+    },
+});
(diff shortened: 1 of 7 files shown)


src/schema/matchmaking/cancelled.ts
```diff
@@ -0,0 +1,15 @@
...
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description:
+        "Server may send this event at any point when the user is queuing to indicate that the user has been booted out the matchmaking system.",
+    event: {
+        data: Type.Object({
+            reason: Type.String(),
```
Jazcash, line 12: I'd like to try and keep properties like this as explicit as possible where we can, unless it really doesn't make sense. For example, here I would change `Type.String()` to `Type.Union([ Type.Literal("intentional"), Type.Literal("party_user_left"), Type.Literal("server_playlists_changed") ])`. If we need to add or remove reasons later we can easily do so, but this approach improves the typing for protocol consumers and makes it easy for client devs to see what possible reasons to expect here.

Jazcash (approved):

src/schema/matchmaking/cancelled.ts
p2004a: And for any free form text there can be additional details field as in responses

geekingfrog: Amended the commit to hardcode some reasons.

PR #35, opened 2024-09-22
Add matchmaking cancelled timeout reason

geekingfrog:
Add a new `cancelled` reason for when a player fails to ready up in time.
Leaving afk players in matchmaking queues could introduce "poison pills": some match would get prepared to only repeatedly fails. So it's better to just boot them out.

### DIFF
--- a/src/schema/matchmaking/cancelled.ts
+++ b/src/schema/matchmaking/cancelled.ts
@@ -10,7 +10,7 @@ export default defineEndpoint({
         "Server may send this event at any point when the user is queuing to indicate that the user has been booted out the matchmaking system.",
     event: {
         data: Type.Object({
-            reason: UnionEnum(["intentional", "server_error", "party_user_left"]),
+            reason: UnionEnum(["intentional", "server_error", "party_user_left", "ready_timeout"]),
         }),
     },
 });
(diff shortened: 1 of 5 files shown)


src/schema/matchmaking/cancelled.ts
```diff
@@ -10,7 +10,14 @@ export default defineEndpoint({
...
     event: {
         data: Type.Object({
-            reason: UnionEnum(["intentional", "server_error", "party_user_left"]),
+            // * intentional: the player left matchmaking
+            // * server_error: something bad happended and the server couldn't maintain
+            //   state so it booted the player out. Retrying may fix the issue.
+            // * party_user_left: a member of the player's party left matchmaking, thus
+            //   forcing the entire party to withdraw
+            // * ready_timeout: The player failed to accept a match within the given
+            //   time window and is thus removed from the matchmaking system.
```
p2004a, line 19: That's not the best place for this documentation, because it's not unexpected to read this source when wanting to learn about protocol. So that fits in the description field, or in the markdown docs for matchmaking.

geekingfrog: I put that in the matchmaking readme and amended the commit

PR #36, opened 2024-09-23
Remove matchmaking/readyUpdate event

geekingfrog:
It is redundant with foundUpdate, and the client should already have the expected number of player for each queues.


p2004a:
Ah, missed https://discord.com/channels/549281623154229250/943582636679520256/1284810222182862910

PR #37, opened 2024-11-18
Remove formating constraint on user id

geekingfrog:
No need to require a UUID, it's treated as an opaque string


p2004a:
I've pushed update to autohost with this PR included

PR #39, opened 2024-11-26
Add more info about connection termination

geekingfrog:
Capture the discussion from https://github.com/beyond-all-reason/tachyon/pull/38 in the doc.
If that's how we should go, #38 can be abandoned.

### DIFF
--- a/docs/connection.md
+++ b/docs/connection.md
@@ -35,6 +35,15 @@ The server must use the WebSocket built-in ping/pong mechanism to keep the conne
 
 The frequency of pings must be at least 1 every 10 seconds, and preferably randomized to some degree.
 
+## Connection termination
+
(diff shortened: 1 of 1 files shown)


docs/connection.md
```diff
@@ -35,6 +35,15 @@ The server must use the WebSocket built-in ping/pong mechanism to keep the conne
 
 The frequency of pings must be at least 1 every 10 seconds, and preferably randomized to some degree.
 
+## Connection termination
+
+In case something goes wrong and the connection has to be terminated, the peer should
+close the connection with the code `1008` for [policy violation](https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4).
```
p2004a, line 41: Unless some other codes are more appropriate, we can define it in the future.

p2004a (approved):

PR #42, opened 2025-01-26
Messaging spec

geekingfrog:
An attempt to define the specs for chats. I think we'll add a few other source/target like `lobby` or `party`.

(also note, the CI on this project does **not** like jujutsu created commit :/ )


src/schema/messaging/send.ts
```diff
@@ -0,0 +1,26 @@
...
+                }),
+            ]),
+            // somewhat arbitrary maxLength, but we need one
+            message: Type.String({ maxLength: 512 }),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "message_too_long" },
+        { status: "failed", reason: "not_connected" },
```
p2004a, line 24: What `not_connected` means? That the player on the other end is no available?

src/schema/messaging/send.ts
```diff
@@ -0,0 +1,26 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
```
p2004a, line 4: For reconnect, the client should just send message again and assumption is that message copies are not a big deal?

src/schema/messaging/received.ts
```diff
@@ -0,0 +1,19 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "Notify the player a message has been received",
```
p2004a, line 8: Received as in, we send message out, and get notification that the other player received it? If there were multiple messages send out, which one?

p2004a (commented):
How recipient receives the actual message?

geekingfrog:
> How recipient receives the actual message?

omg, I forgot the `message` property on the `messaging/received` event -_-"
I have amended the commit to correct that. Also expanded the readme to explain more, let me know if that makes sense.

src/schema/messaging/send.ts
geekingfrog: not sure I follow, which reconnect? If the client never received the response to `messaging/send` then yes, it should send the message again and that may result in duplicate, but that's not a big deal indeed.

I haven't really put much thought into guaranteeing end-to-end delivery of messages, do you think we need that?

p2004a: Agreeing on fire-and-forget/at-least-once/exactly-once is good idea. I believe at-least-once would be good to have, but if you want we can also start with what you purpose and rework later.

p2004a: I think the easiest way for it to be at-least-once is make it a server->client request, not event, and retry in the server in the "session" process (not the TCP connection one). To allow making it exactly-once I guess you could add a increasing id/time-stamp to the message and let the recipients client deduplicate.

p2004a: (I assume we do not want to have a history or anything like that yet)

geekingfrog: ok, let's make the `server -> client` part also a request so that the server knows when a message is safe to delete. At least the proto will support at-least-once delivery, that would be a pain to change later.

geekingfrog: Although, the at least once delivery is going to be problematic for multicast messages, typically for parties and lobbies. I think it will add a ton of complexity there :/

src/schema/messaging/received.ts
```diff
@@ -15,7 +16,7 @@ export default defineEndpoint({
                     player_id: Type.String(),
                 }),
             ]),
+            marker: Type.Ref(unixTime),
```
p2004a, line 19: Didn't you intend to use timeMarker here?

src/schema/messaging/received.ts
```diff
@@ -15,7 +16,7 @@ export default defineEndpoint({
                     player_id: Type.String(),
                 }),
             ]),
```
p2004a, line 18: Should the message also contain time of the message for presentation in UI? If it won't I will assume that lobby will just use marker for that, and it's not intended for that if it's supposed to be opaque.

src/schema/messaging/subscribeReceived.ts
```diff
@@ -0,0 +1,53 @@
...
+    description: "Ask the server to send events for relevant messages",
+    request: {
+        data: Type.Object({
+            source: Type.Union([
+                Type.Object({
+                    type: Type.Literal("player"),
+                    player_id: Type.String(),
+                }),
+            ]),
+            since: Type.Optional(
```
p2004a, line 18: If it's optional, what is the default behavior?

src/schema/messaging/subscribeReceived.ts
```diff
@@ -0,0 +1,53 @@
...
+                ])
+            ),
+        }),
+    },
+    response: [
+        {
+            status: "success",
+            data: Type.Optional(
+                Type.Object({
+                    history: Type.Array(
```
p2004a, line 39: Instead of returning history like that, have you considered just push history via events? Basically you subscribe to the past, and get past events via regular means.

src/schema/messaging/received.ts
```diff
@@ -15,7 +16,7 @@ export default defineEndpoint({
                     player_id: Type.String(),
```
p2004a, line 16: We use `src/schema/definitions/userId.ts` everywhere else, maybe it should be dropped and just a string, but at least the name of property would be nice to name `userId`. The same below.

src/schema/messaging/subscribeReceived.ts
```diff
@@ -0,0 +1,53 @@
...
+                        Type.Object({
+                            message: Type.String(),
+                            marker: Type.Ref(unixTime),
+                        })
+                    ),
+                    has_missed_messages: Type.Boolean({
```
p2004a, line 45: we use camel case for all properties, and snake case for string literals

src/schema/messaging/received.ts
```diff
@@ -1,12 +1,13 @@
 import { Type } from "@sinclair/typebox";
```
p2004a, line 1: Hmm, I'm not sure about naming, maybe just `message`? But really not strong opinion, feel free to just resolve this comment.

src/schema/messaging/subscribeReceived.ts
bcdrme, line 1: I'd drop everything that is in here 

So basically, as a client, you are either subscribed or you're not. Anything that is being sent to you while unsubscribed should be lost forever since :

> There is no support to store messages to be then delivered later. In short, we're not building
whatsapp or discord.

Given that, 
- there's no point in allowing complex fetching behavior such as `since` imho.
- there's also no such thing as a `history`. If there's some sort of backlog of unsent messages in the process because of a short websocket disconnection, just send them all one after the other.

And finally, I don't see the point in having `hasMissedMessages`, it doesn't seem useful to me from a client perspective.

src/schema/messaging/received.ts
bcdrme: The timestamp of each message should be set by the server so we don't rely on client implementation. And I think it should be part of the data object yes.

src/schema/messaging/received.ts
bcdrme: I'd suggest `chat`.

src/schema/messaging/subscribeReceived.ts
p2004a: I disagree, this stands in full contradiction to us wanting to provide reconnect support.

"permament history" != "session level state"

If we want to support transparent websocket reconnection, to which we agreed multiple times in the past, this is required.

bcdrme: yeah that's what I meant by 

> If there's some sort of backlog of unsent messages in the process because of a short websocket disconnection, just send them all one after the other upon re-connection.

p2004a: You can't do that, you don't know if websocket message was devlivered or not.

bcdrme: Then we need to ack everything. You'd have the same issue with a request for history or whatever. How would you know if it has been received by the client ? How would the client know it has missed messages and need to ask for a history again ?

p2004a: Not dropping data while connection is open is a guarantee of TCP connection.

You can do explicit acks for everything (convert all communication to request response effectively) but that just adds additional unnecessary traffic on the network. With acks, you also need to deal instead with deduplication of messages on client side, because ack might be missed by server when connection drops, so then it will resend the same message again.

p2004a: To expand for 
> How would the client know it has missed messages and need to ask for a history again ?

because maybe I've misunderstand what you meant: That's what "since" call is for. You pass in the marker of the last ID you've seen on client.

src/schema/messaging/received.ts
geekingfrog: added a timestamp

src/schema/messaging/subscribeReceived.ts
geekingfrog: I didn't find a way to mark a default option, so I put it in the readme.

p2004a: I've pushed commit that sets the default

p2004a (approved):

PR #44, opened 2025-02-02
Add user/self endpoint and modifications to related endpoints

bcdrme:
- **Add user/self endpoint. Update user/updated accordingly.**
- **Add a mean to subscribe to a set of userIds.**
- **Add current battle information to privateUser state for reconnection purposes.**


src/schema/user/subscribeUpdates.ts
```diff
@@ -0,0 +1,24 @@
...
+        data: Type.Object({
+            userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
+        }),
+    },
+    response: [
+        {
+            status: "success",
+        },
+        {
+            status: "failed",
```
p2004a, line 20: "failed" doesn't mean anything, note that every request/response already by default has

```
                        "internal_error",
                        "unauthorized",
                        "invalid_request",
                        "command_unimplemented"
```

src/schema/user/self.ts
```diff
@@ -0,0 +1,18 @@
...
+import { userId } from "@/schema/definitions/userId";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description:
+        "Sent by the server to inform the client of its own user state. This event should be sent to a user when they login.",
+    event: {
+        data: Type.Object({
+            user: Type.Deref(privateUser, [user, userId]),
```
p2004a, line 15: As discussed, it it's always full object, it should be `Ref`

src/schema/definitions/privateBattle.ts
```diff
@@ -0,0 +1,14 @@
+import { Type } from "@sinclair/typebox";
```
p2004a, line 1: Maybe also use it in the `battle/start`? It's the same type.

src/schema/user/subscribeUpdates.ts
```diff
@@ -0,0 +1,24 @@
...
+            reason: "limit_reached",
```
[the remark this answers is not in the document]
bcdrme: So I added that because I assume there would be a server limit to how many users you can be subscribed to. 
It would force client to unsubscribe to whatever users are not relevant anymore (unsubscribe is lacking in the protocol btw I can add that)

src/schema/definitions/privateBattle.ts
bcdrme: yes that's already the case 

https://github.com/beyond-all-reason/tachyon/pull/44/files#diff-3a0737d6c16b5000b4c4c2158b5fd2d9be6e5eeefa9af3d4bc140059880dc415:~:text=data%3A%20privateBattle,()%2C

src/schema/user/subscribeUpdates.ts
bcdrme: but this is what we have done for other parts of the protocol though

failed + reason

src/schema/user/self.ts
bcdrme: Changed

src/schema/user/subscribeUpdates.ts
p2004a: Wait, I had massive brain fart here 🤦‍♂️. All good, it's the "type" filed xd, not reason field xd.

src/schema/definitions/privateBattle.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const privateBattle = Type.Object(
+    {
+        username: Type.String(),
+        password: Type.String(),
+        ip: Type.String(),
+        port: Type.Number(),
```
geekingfrog, line 8: We should add then engine and game full names/version used for this battle. This way the client has a way to automatically reconnect, and it also eliminates any potential race conditions around assets sync.

src/schema/user/subscribeUpdates.ts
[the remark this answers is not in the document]
geekingfrog: +1 for `subscription_limit_reached`, and yes, there's likely going to be some limit enforced server side, but likely higher than 200. Especially because some people have 300+ friends.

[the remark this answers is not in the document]
bcdrme: We can't be allowing 300+ friends that's not practical. You'd have to send keep track and send so many things, there's no way. 
That's also an issue client side. Moving this discussion on Discord.

src/schema/definitions/privateBattle.ts
p2004a: Ah, good, let's just make sure that it's `$ref` and not inline and we are good

src/schema/definitions/privateBattle.ts
bcdrme: I'm gonna also add map actually, in case somehow we need to redownload stuff

PR #45, opened 2025-02-02
Add `autohost/installEngine` and `availableEngines` field.

p2004a:


src/schema/autohost/installEngine.ts
```diff
@@ -0,0 +1,33 @@
...
+                    {
+                        version: "2025.01.5",
+                    },
+                ],
+            }
+        ),
+    },
+    response: [
+        {
+            status: "success",
```
geekingfrog, line 25: What does success mean here? request received, engine installed, something else?

p2004a: Installed, I've added the description.

geekingfrog: How long do you reckon that operation could take? I’m thinking I may need a higher timeout for response on that one maybe?

p2004a: Hard to tell, I will have some internal timeouts and retries, they will be probably max ~5min, because download from GitHub can be very very slow in some cases. In normal situations it should be < 15s

Would you prefer to instead have just "trigger download" or something like that?

I don't like long running rpcs too...

p2004a: If you want we can add `deadline` argument and I will ensure return before deadline is reached.

geekingfrog: not sure about the deadline bit, seems it’s just going to add complexity for not much benefit.
If autohost gets another request to install an engine while it’s already dealing with the same request, what should happen?
I can increase the timeout and do some retries, and in this case it would be good to know how to handle that use case.

p2004a: If you send another request for the engine version that was already requested to be installed via a previous request, I will dedup it internally.

So it's really a question for you:
- Do you want instant response
- Or do you want possibly long response but also potential report with error (not sure of you need it anyway)

The available versions are still distributed via the status update, so it's effectively a question if you want/need to track or something I dunning.

For me, for autohost implementation, it doesn't matter at all which option you pick.

geekingfrog: I'd say let’s make this a request to do things in the background, and "success" means the autohost acknoledged it, but may not have completed it.

p2004a: Okey, pushed the change.

geekingfrog (approved):

PR #51, opened 2025-02-28
Friendlist

geekingfrog:
The messages for friend list features.


src/schema/friend/request_cancelled.ts
Jazcash, line 1: This is identical to `request_rejected`, could this info instead be communicated in that event with an additional property, e.g. `data: { reason: "cancelled" }`?

src/schema/friend/list.ts
```diff
@@ -0,0 +1,32 @@
...
+            outgoing_pending_requests: Type.Array(
+                Type.Object({
+                    to: Type.Ref(userId),
+                })
+            ),
+            incoming_pending_requests: Type.Array(
+                Type.Object({
+                    from: Type.Ref(userId),
+                })
+            ),
```
Jazcash, line 28: As these are single-property objects, I'd recommend not having an object at all and instead just `Type.Array(Type.Ref(userId))`

Jazcash (commented):
I think the command names (taken from the file name) should be camelCase as that's how the rest of the schema is formatted.

src/schema/friend/list.ts
bcdrme: You should have the timestamp in there for sure. We'd want to order the requests from newest to oldest

src/schema/friend/request_cancelled.ts
geekingfrog: I think I prefer having them separate. Because the originator is different, in the case of `reject`, it's the person who received the request that is acting, and for `cancel` it's the person who sent the request who's the actor.
If we merge the two, I'm worried that it's going to be ambiguous about which request should be used when.

Jazcash (approved):

src/schema/friend/list.ts
```diff
@@ -0,0 +1,34 @@
...
+    description: "Retrieve the status of your friendlist",
+    request: {
+        data: Type.Object({
+            friends: Type.Array(
+                Type.Object({
+                    userId: Type.Ref(userId),
+                    addedAt: Type.Ref(unixTime),
+                })
+            ),
+            outgoing_pending_requests: Type.Array(
```
p2004a, line 19: All property names should be camel case. We should figure out a linting rule for this....

PR #53, opened 2025-03-07
Party spec

geekingfrog:


src/schema/party/invite.ts
```diff
@@ -0,0 +1,21 @@
...
+    request: {
+        data: Type.Object({
+            userId: Type.Ref(userId),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "not_in_party" },
+        { status: "failed", reason: "invalid_user" },
+        { status: "failed", reason: "target_already_in_party" },
```
Jazcash, line 19: Am I right in thinking `target_already_in_party` as a possible failed response here indicates it won't be possible to even send a party invite to someone if they're already in a party? Not sure I like that

src/schema/party/README.md
Jazcash, line 1: Imo would be good to make any references to specific commands as actual links, like at the top of https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/matchmaking.md

src/schema/party/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+A player can only be in at most one party at any given time.
+
+Any player in a party can invite any other connected player. When an invite is
+sent, the target player as well as every party member gets an event
+`party/invited`.
+
+Any pending invite can be cancelled by any member in the party with the request
+`party/cancelInvite`. Afterward, an event `party/inviteCancelled` will be sent
+to all party members and invited players.
```
Jazcash, line 14: Invites by one member of the party will be cancellable by other party members? Not necessarily sure I have a problem with it, but it's atypical, and I'd like to hear some reasoning for it.

src/schema/party/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Accepting an invite can be done with `party/acceptInvite`, and will be followed
+by an event `party/memberJoined` sent to all members and players with pending
+invites. Declining is done with `party/declineInvite` and will be followed by
+`party/inviteDeclined`.
+
+Any member in a party can kick any other member with the request `party/kickMember`.
```
Jazcash, line 21: Similar thought, usually only the party leader can kick people.

src/schema/messaging/send.ts
Jazcash, line 1: Shouldn't the same addition be made to `messaging/received`?

src/schema/party/memberKicked.ts
Jazcash, line 1: Same thought that I gave in the other review, there's a lot of duplicate commands that are given their own schema for the sake of semantic difference, rather than practical difference. In this case, I personally would prefer only `party/memberLeft` with a property like:
```
reason: Type.Union([
    Type.Literal("left"),
    Type.Literal("kicked"),
    Type.Literal("disconnected")
])
```
I think this keeps the protocol thinner, reduces the amount of duplicated schema, and potentially narrows consumer logic into a simpler place, because they don't have to be aware of and account for multiple commands that ultimately will likely all handle the same way.

Jazcash (changes requested):
In the description for `party/create`, there's reference to party leader, but this isn't mentioned anywhere else. I'd expect to see references like `partyLeader: Type.Ref(userId)` in various places. I think this will also need a command for passing leadership, and an event for leadership being transferred.

I think there should be a timeout for invitations, configurable by the server. Probably in `party/invited`, and then `timed_out` as a failure reason in `inviteCancelled/inviteDeclined`.

I see a couple of responses contain timestamps like `invitedAt` or `joinedAt` , but I'm wondering if these are present to solve some practical problem or just if they might potentially be needed. If they are solving a problem, could that problem also be solved if every command included something like `sentAt: Type.Ref(unixTime)` in its root?

src/schema/party/get.ts
Jazcash, line 1: Is this needed? It seems like all party members are informed of info like other players being invited, or invitations being declined etc. Would the consumers not already be fully aware of the party state at all times anyway?

geekingfrog: I think you need that when you get an invite, to get the state of the party, like existing members and invites.
Maybe we can put this info into `party/invited` event. I think it's also useful for clients that reconnects, the `user/self` has a partyId that can be used with `party/list`. If we remove the `list`, then we would need to put all the data inside `user/self` as well.

src/schema/party/invite.ts
geekingfrog: Yeah you’re right, I think this error should probably be for accepting an invitation when already in a party, will change.

src/schema/party/memberKicked.ts
geekingfrog: I like this suggestion, although I think we can remove `disconnected`. @bcdrme any thoughts on that?

geekingfrog:
I think I would remove all mention of party leader. The idea is that it simplify the implementation (it’s not very complicated, but still), and that parties are supposed to be a group of friends, so "moderation features" should be irrelevant here.

Otherwise what would the semantics be?
* only leader can start matchmaking/join something? We had a discussion and it may work, but it's not ideal either (if leader temporarily afk for example)
* only the sender of an invite can cancel it?
* only leader can kick players out of party?
* when leader leave, oldest remaining member is promoted?
* do we need commands to transfer leadership?

src/schema/party/README.md
bcdrme: Not usually, it depends on the game. Me personally I'm comfortable with everyone in the party being able to invite / kick / start matchmaking. See this discussion https://discord.com/channels/549281623154229250/943582636679520256/1346419896006737962

src/schema/party/memberKicked.ts
bcdrme: I think memberLeft with a reason is good yeah.
For `disconnected`, I think it's kind of the same thing as `left`. We can drop it imho

bcdrme:
> I think I would remove all mention of party leader. The idea is that it simplify the implementation (it’s not very complicated, but still), and that parties are supposed to be a group of friends, so "moderation features" should be irrelevant here.

I agree with this statement. We don't need that complication, especially if we have an `invite only` mode that you can toggle on the party.

src/schema/party/memberKicked.ts
geekingfrog: Ok, done, I removed `party/memberKicked` and added a reason to the `party/left` message.

src/schema/party/README.md
Jazcash: I say usually because I've played quite a lot of different online matchmaking games with party support and I think in all of them anybody can invite people to the party, but only the leader can control what the party queues for, and I think in most cases only the party leader can kick people also. Again, I don't necessarily have a problem with everybody in the party effectively being a party admin, but I think it warrants some more discussion as to why other games don't do it this way and if there's any potential problems with doing it this way.

Jazcash:
Ok, maybe it's fine not having party leader. As I mentioned I think this way is pretty unusual, but no obvious reasons as to why not to do it this way come to mind.

src/schema/party/get.ts
Jazcash: I do think I'd prefer to see this info given in existing places like in the `success` response of `acceptInvite`, or `invited` as you say. These kind of `get`/`info` requests seem a bit arbitrary in exactly when they should be needed, and I think adding the info to existing responses/events makes it easier for consumers, rather than being expected to send this request themselves at specific times.

Reconnects sound like they might be complex, and tbh I'm sceptical if we should ever care about them for this early version of the protocol, at least for parties.

L-e-x-o-n:
I think it makes sense that parties have limited size, if not set otherwise then a party size of 8 should be the max (being the largest possible currently supported team size). With that in mind, the invite request would need an additional fail reason of `party_full`.
How to handle existing invites to full parties (can happen if multiple invites are sent, one of which is accepted causing the party to reach max capacity while other invites are still active)? I think the `acceptInvite` request should fail in this case with `party_full`. Alternatively, sending an `inviteCancelled` for the remaining invites could be considered.

L-e-x-o-n:
On the other hand, how does a party end? It should probably have a minimum size (should it be one or 2 members?) and stops existing when it's member count goes below.
What should happen when the last or next to last player leaves the party,?
Concerning the invited players specifically: do we want some kind of party disbanded event to inform the invitees that the party no longer exists, should we just send cancel invites to everyone invited or should we rely on invited clients tracking party size by member joined/left and the `invalid_party` response to accepting such invites?

L-e-x-o-n:
`party/leave` should probably have `not_in_party` fail reason. `party/kickMember` has it but I think it's for the player you are trying to kick and not yourself. Additionally, it might make sense to not allow kicking yourself which would then require another fail reason.

geekingfrog:
> With that in mind, the invite request would need an additional fail reason of party_full.

I’ve added a `party_full` failure and some notes in the readme. I think for simplicity sake, the size of the party is the number of members + pending invite, this way there is no case where accepting an invite fails because already full.

> On the other hand, how does a party end? It should probably have a minimum size (should it be one or 2 members?) and stops existing when it's member count goes below.
What should happen when the last or next to last player leaves the party,?

I clarified that in the readme: when last member leaves, party is disbanded, and all pending invited players receive a `inviteCancelled` event.

> `party/leave` should probably have not_in_party fail reason.

I don't think it makes sense. What would a client do with that information? most likely, it'll ignore it since the goal is achieved. Making it a no-op also makes it idempotent which is always a nice property to have.

> ` party/kickMember` has it but I think it's for the player you are trying to kick and not yourself.

I've added a new failure code for this: `target_not_in_party`

>  Additionally, it might make sense to not allow kicking yourself which would then require another fail reason.

I think that's unnecessary. It's stupid but doesn't really contradict anything. Allowing it would just be simpler I believe.

src/schema/party/get.ts
geekingfrog: Ok, I’ve removed the `party/get` message, and included the party state in `invited`, `memberJoined` and `user/self` events.

L-e-x-o-n (approved):

src/schema/party/get.ts
Jazcash: Maybe I'm looking at the wrong thing but I still `party/get` in the files changed

src/schema/party/README.md
geekingfrog: I am pretty sure we're going to end up implementing something similar to what you're describing. But I also think we can add that later. At the protocol level, it's merely going to be a few more error reasons and a data point to indicate who's the leader, and the implementation shouldn't be too hard either.

src/schema/party/get.ts
geekingfrog: Not sure what happened, it seems I didn't push the commit. There should be a new commit now with these changes

Jazcash (approved):

src/schema/party/memberJoined.ts
```diff
@@ -0,0 +1,21 @@
...
+    source: "server",
+    target: "user",
+    description: "New player joined the party (accepted an invite)",
+    event: {
+        data: Type.Composite([
+            Type.Object({
+                userId: Type.Ref(userId),
+                joinedAt: Type.Ref(unixTime),
+            }),
+            partyState,
```
p2004a, line 18: Not ref?

src/schema/party/invited.ts
```diff
@@ -0,0 +1,22 @@
...
+    target: "user",
+    description:
+        "A player has been invited to the party. Sent to the invited player and all party members.",
+    event: {
+        data: Type.Composite([
+            Type.Object({
+                invitedUserId: Type.Ref(userId),
+                invitedAt: Type.Ref(unixTime),
+            }),
+            partyState,
```
p2004a, line 19: Same here, not ref?

src/schema/party/README.md
```diff
@@ -0,0 +1,35 @@
...
+and the event [party/left](#left) will follow for all remaining members.
+
+# Limitations and expansions
+
+As time of writing, this spec doesn't have any concept of leader, any party member can do all actions.
+Maybe in the future, we may want to restrict some things, like only leader can join games/matchmaking
+or invite/cancel invites.
+
+We also discussed a concept of public parties, where anyone could join. This may be a future extension
+but is out of scope for now.
```
p2004a, line 35: How is reconnect handled? How those events are resend etc?

src/schema/party/memberJoined.ts
geekingfrog: So I tried
```typescript
data: Type.Composite([
    Type.Object({
        userId: Type.Ref(userId),
        joinedAt: Type.Ref(unixTime),
    }),
    Type.Ref(partyState),
]),
```

but it produced the following schema:

```json
        "data": {
            "title": "PartyMemberJoinedEventData",
            "type": "object",
            "properties": {
                "userId": { "$ref": "../../definitions/userId.json" },
                "joinedAt": { "$ref": "../../definitions/unixTime.json" }
            },
            "required": ["userId", "joinedAt"]
        }
```
where there is no partyState at all. Did you have something else in mind?

src/schema/party/README.md
geekingfrog: I am not sure reconnections make sense for parties as this is a purely online system. When you disconnect, you leave the party automatically and any pending invite is invalidated.
I think supporting offline players in party would add a ton of complexity but also be rather confusing since a party is something meant to play with others, and you can't do that with offline players.

p2004a: Nothing should be tied to the state of websocket connection. That's the point of reconnects, stuff like server restart, client disconnecting and connecting because they wifi flickered/they changed access point needs to be transparent.

geekingfrog: There's already a concept of session for player. If the connection dies but they come back soon enough then it's as if they never left.
If that lasts too long, then they are considered gone.

p2004a: And that's the question: how does session resumption work with parties API? What happens with all the events?

src/schema/party/memberJoined.ts
Jazcash: I think you might want `Type.Intersect` instead of `Type.Composite`, see https://github.com/beyond-all-reason/tachyon/blob/master/src/schema/definitions/privateUser.ts

src/schema/party/README.md
Jazcash: Is this critical for first draft / MVP? I think my recommendation would be to do them after we have something functioning to make it easier to identify the problems with intermittent connections and a better understanding of how to go about solving them.

p2004a: In many places it impacts protocol design on a fundamental level, it impacts implementation on both client and server on fundamental level, it's not something that can be trivially bolted on top.

Looking at current messages, because party state is part of private user, which is send on connect via `user/self`, losing events related to the joined party seems ok. Client will just get latest state of the party on reconnect.

But looks like there isn't any ability to see the list of invites that player received, but not responded yet, which will be required for reconnect.

What seems to be missing

src/schema/party/memberJoined.ts
p2004a: Okey, let me step back actually.

Why do we want to put party state in this message at all? Client will have current state of parties anyway, it can just add element. If we want to push the party state, instead we could drop the `party/memberJoined`, `party/memberLeft` and just have `party/update` that pushes whole party state because it's small anyway.

p2004a:
Looking back on the previous conversation, I would also like to touch on custom error reasons: I don't believe it's good idea to have custom error codes for every possible failure case. For many cases it's just IMHO unnecessary, it might be just more cases to cover for client and so on. Custom errors are important for cases where we really need client to be able to distinguish between them and take a different action.

Every single request can always get in the response: `unauthorized`, `internal_error`, `invalid_request`, `command_unimplemented`. If the data passed by the client doesn't make sense in some way (e.g. you try to kick yourself, or even you send leave perty when not in party) it's often reasonable to just answer `invalid_request`.

Right now we have a flat space of error codes, and let's see how ergonomics of it looks in practice, but there is good reasons why https://en.wikipedia.org/wiki/List_of_HTTP_status_codes and e.g. https://grpc.io/docs/guides/status-codes/ look the way they look like: it allows for consistent handling, e.g. it allows for standard approach for retries, hedging etc https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L495, https://github.com/grpc/proposal/blob/master/A6-client-retries.md. IMHO this is way less of a concern for Tachyon, and at the moment I don't foresee a need to e.g. group errors into categories, have some more structure around them, but we will see.

src/schema/party/README.md
bcdrme: When you reconnect to your _**not yet dead**_ session, the server should send you your complete user state. IIRC, that's what we talked about for being able to reconnect to an ongoing battle, you'll receive the necessary stuff so you can reconnect. I think we should consider that a default behavior for other stateful systems as well, such as party. 

What I'd suggest is adding to said user state the partyID, and then it's a client concern to resubscribe to the updates. "resubscription" might just be a join party request.

src/schema/party/memberJoined.ts
Jazcash: `user/updated` works like that, but it allows partial data, so the server can send only the properties that changed, or the full state if it wants. It does keep the protocol spec lighter, but probably most important thing is consistency.

Jazcash:
Regarding custom failed reasons, I think clients in most cases will appreciate knowing why their request failed, even if there's no visual difference to end-users in some cases, I think it's still very useful for devs.

The example you gave about a user trying to kick themself being met with `invalid_request` I think is correct, because as you say it's impractical to have failure codes for every possible use-case, but this is also the definition of an illegal request in the first place.

As for sending a `party/leave` request when not in a party, I think `invalid_request` is fine, because it's also an illegal request, but ultimately this is a bug on the consumer's side that they should fix. Somewhere they're allowing the sending of `party/leave`, despite not being in a party, and getting a `not_in_party` type response would make that a lot easier to realise. And treating it as a no-op or even success response from the server is problematic because the client dev may never realise there's an issue on their end in the first place.

A benefit of generating strict type defs and using strict mode in the new client is it makes it easy to ensure all types of response are handled, or otherwise fail compilation, which I think is a good thing. If the custom failed reason doesn't matter, i.e. if they should all be handled the same way, then the types of response that do matter can be handled and everything else can be caught in an `else` block.

p2004a:
> Regarding custom failed reasons, I think clients in most cases will appreciate knowing why their request failed, even if there's no visual difference to end-users in some cases, I think it's still very useful for devs.

That's why there is `details` field. All possible failures in many cases can be close to impossible to enumerate in enum.

Client needs to be able to distinguish errors it is expected to handle differently. If many invalid argument type errors are just indication of client bug, there isn't a need to break it down when `details` field contains all the info for developer.

I've brought up already grpc and http as examples, but there are many more, e.g. see how OAuth2 deals with this in similar way: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 or if you want even more enterprise way: https://google.aip.dev/193 that adds a ton of structure and extendability. Basically, there is always some reasonable grouping, and a way to return more details for debugability. It's really a balance.

Jazcash:
Ah, I wasn't aware a details field was added to failure responses (probably should be added [here](https://github.com/beyond-all-reason/tachyon?tab=readme-ov-file#responses)). In that case, I think you're right. Keep `reason` enum limited to cases the client may wish to handle differently, although that might not always be obvious, but we can review them for each response as they're PR'd.

src/schema/party/README.md
geekingfrog: I think we're fine here, because upon connection, the first event the player automatically gets contains the party state, and afterwards they'll get all updates. So even when clients have spotty connection they won't lose state, unless they're gone for too long.

src/schema/party/memberJoined.ts
geekingfrog: So I started to change the spec to use a single `party/updated` event and it raises a few questions:

1. Does this event also replace all the invite events? The party state contains the list of invite, so that's doable.
2. When sending only the state, there's no way to know if a player has been kicked or left on their own accord. Same for cancel vs decline an invite. Is this something we care about?

src/schema/party/README.md
geekingfrog: > But looks like there isn't any ability to see the list of invites that player received, but not responded yet, which will be required for reconnect.

Good point, I'll extend `privateUser` with this data

geekingfrog:
I have removed the various errors code.
I've also pre-emptively merged all events into the all encompassing `party/updated` one, but I'm not sure about some things, see [my previous comment](https://github.com/beyond-all-reason/tachyon/pull/53#discussion_r1997530306).

src/schema/party/memberJoined.ts
Jazcash: 1, 2. Yeah, I think if everything can be done through `updated` then it's fine to drop `invited`.
3. I don't think we care, at least not for a while

As I half-mentioned previously, benefits to this method is it keeps the protocol thinner and makes it easy for the server if it just wants to send the full state every time. The downside is the client now has to compare the new state to the old each time to work out the differences and payloads are larger than they need to be, potentially being an issue if the state model is expanded by a lot. Personally I'm not too fussed which approach we go with, but this way is at least consistent with how `user/updated` currently works.

It's also worth mentioning that the current `party/updated` type _requires_ the full state, rather than `Type.Partial...` as `user/updated` does. Partial states allows the server to send only the properties that changed, allowing payloads to be thinner and less comparison work the client needs to do. But with this approach it'd probably be better to have a separate command that contains the full state (e.g. `invited`), to make sure the client first gets told the full state up front before receiving partial updates.

Honestly though this is kinda optimisation stuff that can probably be done further down the line, I'm fine with it sending the full state each time for now.

PR #57, opened 2025-05-16
Add player id in source message in party

geekingfrog:
Knowing a message is coming from a party is useful, but not enough for a client to be able to display the sender in a meaningful way, oops.


p2004a:
😂

p2004a (approved):

Issue #58, opened 2025-06-12
Send player OS to clients


geekingfrog:
Both the `user/info` response and `user/self` event should have data about that player's skill.
Something like:

```
skill: {mu: float, sigma: float}
```

so that clients can display this information. It's also going to be useful when debugging matchmaking when skill based matching algo is implemented (https://github.com/beyond-all-reason/teiserver/issues/673)

L-e-x-o-n:
`mu` is often referred to as `skill` in other places so it would be better if it is replaced with `rating`.
```
rating: {mu: float, sigma: float}
```

L-e-x-o-n:
Not sure if both mu and sigma should be present, there is some discussion about changing how ratings are handled and displayed to players and it might be beneficial to just send a single number. Personally I would prefer to have internal (used by the rating system, no restrictions or worry about making it visually pleasant) and external ratings (displayed to the players, can't be negative, not shown until first X wins/games etc.). I think it would be better to handle this on the server in this case.

PR #61, opened 2025-06-21
Add player skill

geekingfrog:
Address https://github.com/beyond-all-reason/tachyon/issues/58

Only skill for now, if we want to add more data like `provistional` or whatever, we'll see when we know what we want exactly.


geekingfrog:
I've removed the comment. And also changed `rating` from a scalar to an object, so that it's easier to extend. If we ever want to send the uncertainty, or whatever, having an object will make it easier to extend without bloating the user object.

PR #64, opened 2025-08-02
Fix UserSubscribeUpdatesRequestData typing

bcdrme:
The typing for UserSubscribeUpdatesRequestData and UserUnsubscribeUpdatesRequestData is wrong.

1. UserId shouldn't be redefined.
2. minItems / maxItems messes up the type generation

```
type UserId1 = string;
type UserId2 = string;

interface UserSubscribeUpdatesRequestData {
    /**
     * @minItems 1
     * @maxItems 100
     */
    userIds: [UserId1, ...UserId1[]];
}

interface UserUnsubscribeUpdatesRequestData {
    /**
     * @minItems 1
     * @maxItems 100
     */
    userIds: [UserId2, ...UserId2[]];
}
```

after:
```
interface UserSubscribeUpdatesRequestData {
    userIds: UserId[];
}

interface UserUnsubscribeUpdatesRequestData {
    userIds: UserId[];
}
```

### DIFF
--- a/src/schema/user/subscribeUpdates.ts
+++ b/src/schema/user/subscribeUpdates.ts
@@ -10,7 +10,7 @@ export default defineEndpoint({
         "Ask the server to send updates about theses users. A maximum of 100 userIds can be subscribed to at any given time.",
     request: {
         data: Type.Object({
-            userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
+            userIds: Type.Array(Type.Ref(userId), { minItems: 1, maxItems: 100 }),
         }),
     },
     response: [
--- a/src/schema/user/subscribeUpdates.ts
+++ b/src/schema/user/subscribeUpdates.ts
@@ -6,7 +6,8 @@ import { userId } from "@/schema/definitions/userId";
 export default defineEndpoint({
     source: "user",
     target: "server",
-    description: "Ask the server to send updates about theses users.",
+    description:
+        "Ask the server to send updates about theses users. A maximum of 100 userIds can be subscribed to at any given time.",
     request: {
         data: Type.Object({
             userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
--- a/src/schema/user/unsubscribeUpdates.ts
+++ b/src/schema/user/unsubscribeUpdates.ts
@@ -10,7 +10,7 @@ export default defineEndpoint({
         "Ask the server to stop sending user updates for the given set of userId. This should always succeed.",
     request: {
         data: Type.Object({
-            userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
+            userIds: Type.Array(Type.Ref(userId)),
         }),
     },
     response: [
--- a/src/schema/user/unsubscribeUpdates.ts
+++ b/src/schema/user/unsubscribeUpdates.ts
@@ -10,7 +10,7 @@ export default defineEndpoint({
         "Ask the server to stop sending user updates for the given set of userId. This should always succeed.",
     request: {
         data: Type.Object({
-            userIds: Type.Array(Type.Ref(userId)),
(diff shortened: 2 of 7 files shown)


bcdrme:
That's weird... :thinking:

bcdrme:
> I don't see `{ minItems: 1, maxItems: 100 }` messing up types. I've added it back locally and it worked fine.

Are you sure @p2004a ? Cause I can repro 100% of the time. There's a quirk, I have to delete `dist` folder to see changes.

p2004a:
> Are you sure @p2004a ?

`npm run build:clean` results in no dupes. To be sure I've done new clone, tested by only adding `Type.Ref`, no dupes.

p2004a:
Ahhh, `userIds: [UserId, ...UserId[]];` is correct type for `{ minItems: 1, maxItems: 100 }`. What's wrong about it? [It works as expected](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAsglgOwIwBkA8AVAfFAvFAbQwBooA6CjAgXWoG4AoBgMwFcEBjYOAewSjgBneMnTYAFAEMAXFCrUAlLMkDBsRKkw4A3gyhQAThGCsD-SWQA2EBAHNgACyg4ADIwC+TNp258ozKVkRTUFgA0RbLAUobU8GAHp4-3EaaMSAWnSoCAMDHgMoACIAZR5TDmgHSTUXbOsAWxtgcUFogCNWYChgSQNbY0MIAEdWOCM1JDJCrxSAcklZxShEqB4AawBCGYJ52dJZtsW0pPWthg4+UO7ZUPC7GjxCXf3D+iYVgOBo5aTMqBAIGBSDk8gZgQAPSBcCAAEyYcGYUHEQmCKHEX2iun0nwUHiAA).

p2004a:
But, if you don't like it and there is some good argument not to do it, https://github.com/bcherny/json-schema-to-typescript has `ignoreMinAndMaxItems` option:

> Ignore maxItems and minItems for array types, preventing tuples being generated.

bcdrme:
Oh my... I'd need a type and a function to check on the length of the array for each case we have in protocol. It doesn't help us anyway in this particular case because the limit of 100 subscriptions is for the whole session, not just for one call. I'd still need to handle the backend error response regardless.

That's the opposite of simple imo. I officially do not like it.

p2004a:
Sure, I agree *ergonomics* of those types is subpar. Please amend the PR to switch the type generator option.

bcdrme:
Done. Please note that other parts of the protocol set `minItems: 1`. That's potentially breaking.
image

src/schema/user/unsubscribeUpdates.ts
```diff
@@ -10,7 +10,7 @@ export default defineEndpoint({
         "Ask the server to stop sending user updates for the given set of userId. This should always succeed.",
     request: {
         data: Type.Object({
-            userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
```
p2004a, line 13: Don't drop it

src/schema/user/subscribeUpdates.ts
```diff
@@ -6,7 +6,8 @@ import { userId } from "@/schema/definitions/userId";
 export default defineEndpoint({
     source: "user",
     target: "server",
-    description: "Ask the server to send updates about theses users.",
+    description:
+        "Ask the server to send updates about theses users. A maximum of 100 userIds can be subscribed to at any given time.",
     request: {
         data: Type.Object({
             userIds: Type.Array(userId, { minItems: 1, maxItems: 100 }),
```
p2004a, line 13: Please bring back the Type.Ref fix

p2004a (changes requested):

p2004a:
> that's potentially a breaking change

Looking at what pops up, it should be fine

p2004a (approved):

PR #69, opened 2025-09-14
Change lobby/listUpdated and add listReset

geekingfrog:
After some discussion in discord, this is a more consistent approach between the lobby list and the lobby details.
The `listReset` event is required when the server needs to recover some internal state, to avoid ghost lobbies.


docs/schema/lobby.md
```diff
@@ -847,9 +852,112 @@ export interface LobbyLeftEventData {
...
+}
+export interface LobbyOverview {
+    id: string;
+    name: string;
+    playerCount: number;
+    maxPlayerCount: number;
+    mapName: string;
+    engineVersion: string;
+    gameVersion: string;
+    currentBattle: {
```
Hectate, line 951: Shouldn't ``currentBattle`` be optional here too?

geekingfrog: Yes good point, I amended the commit.

src/schema/lobby/listReset.ts
```diff
@@ -0,0 +1,15 @@
...
+import { defineEndpoint } from "@/generator-helpers.js";
+import { Nullable } from "@/typebox-utils";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "Sent by the server to give the client the full list of all lobbies",
+    event: {
+        data: Type.Object({
+            lobbies: Type.Record(Type.String(), Nullable(Type.Ref("lobbyOverview"))),
```
p2004a, line 12: Why is lobbyOverview nullable?

PR #76, opened 2025-10-06
Split members into players and spectators

geekingfrog:
Because the modelling with union type doesn't work well with json patch approach, when a user changes type, from player to spectator for example.
By having two different objects for players and spectators, moving a user from player to spec would instead use the following patch:

```
{
    "players": {"123" => null},
    "spectators": {"123" => {"id": "123"}}
}
```


src/schema/lobby/updated.ts
```diff
@@ -49,25 +49,29 @@ export default defineEndpoint({
...
-                                player: Type.Optional(Type.String()),
-                            }),
-                            Type.Object({
-                                type: Type.Const("spec"),
-                                id: Type.Ref("userId"),
-                                joinQueuePosition: Type.Optional(Nullable(Type.Number())),
-                            }),
-                        ])
+                        Type.Object({
+                            type: Type.Const("player"),
```
p2004a, line 58: Missed dropping the const

PR #77, opened 2025-10-06
wip mod support

tetrisface:
[bar-lobby PR](https://github.com/beyond-all-reason/bar-lobby/pull/487)
[teiserver PR](https://github.com/beyond-all-reason/teiserver/pull/820)


Hectate:
As noted in the `bar-lobby` PR, this isn't something that is going to be addressed immediately. If we do get to the point that modding is being evaluated, the repositories will likely be in a very different state. Accordingly, I'm going to close this PR. The request for modding features can be revisited in the future, of course.

PR #79, opened 2025-10-16
Update lobbies

geekingfrog:
For now, it seems we can do everything with a single `update` request.

### DIFF
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -2934,6 +2934,9 @@ Sent by the server whenever something in the lobby changes. Uses json patch (RFC
                                     "type": "object",
                                     "properties": {
                                         "id": { "type": "string" },
+                                        "hostUserId": {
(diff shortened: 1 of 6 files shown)


geekingfrog:
> Making sure I understand; the `AllyTeamConfig` is optional, however if sent the _entire_ config has to be included. I can't just pick one team and adjust their startbox, for example. I'm not against requiring the entire config, I actually think that's safer - just checking to make sure my understanding matches yours.

Yes, clients need to send the entire ally team config. The object is never going to be big, and it's a lot easier to deal with when it's not a partial update. The logic to figure out the required changes in the lobby is going to be gnarly enough.

> It's worth mentioning, the `AllyTeamConfig` interface appears to be the same as `lobby/create` uses. Surely it could be recycled to be used in both places?

I don't understand what you mean. The schema are generated from a shared `ref` so they are indeed the same.

p2004a (approved):

PR #85, opened 2025-12-18
Extension of the protocol for a clan system

Rimek86:
This PR includes protocol extensions for my planned clan system.
Please see clan.md for detailed description.
 _(I added to uml Use-Case-Diagramms and a overall description. I hope this is good for you!?)_

**I look forward to your feedback.**

_Disclaimer: This is my first PR for BAR. I hope i did everything correct. ;-)_ 

NOTE (2025-12-26): Thank you very much for all the feedback. Due to the volume and also the discussions on Discord, I have decided to first collect all requirements centrally and document and share a concept, as it affects not only the protocol but also the server and the client. I will publish the link to the documentation here and on Discord shortly and open it up for discussion, and then resume my work (among other things) on this PR. => Therefore i changed the status of the PR to Draft! => See: https://github.com/Rimek86/BAR-Clan-System-Project

**UPDATE (2026-01-03):  Since this PR is already quite advanced, I have decided to try to complete this very large PR because I don't want to discard the work. I hope I have addressed all the feedback and made some adjustments. I look forward to your feedback again. Thank you in advance.**

### DIFF
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -733,7 +733,7 @@ export interface LobbyCreateOkResponseData {
     bots: {
         [k: string]: {
             id: string;
-            hostUserId: UserId;
+            hostUserId: string;
             allyTeam: string;
             team: string;
             player: string;
@@ -1143,7 +1143,7 @@ export interface LobbyJoinOkResponseData {
     bots: {
         [k: string]: {
             id: string;
-            hostUserId: UserId;
+            hostUserId: string;
             allyTeam: string;
             team: string;
             player: string;
--- a/schema/compiled.json
+++ b/schema/compiled.json
@@ -213,7 +213,8 @@
                                     "type": "string"
                                 },
                                 "hostUserId": {
-                                    "$ref": "#/definitions/userId"
+                                    "$ref": "#/definitions/userId",
+                                    "description": "which player will run the bot. It is the same as the player that added the bot."
                                 },
                                 "allyTeam": { "type": "string" },
                                 "team": { "type": "string" },
--- a/schema/definitions/lobbyDetails.json
+++ b/schema/definitions/lobbyDetails.json
@@ -76,7 +76,10 @@
                             "description": "server assigned unique identifier for the bot",
                             "type": "string"
                         },
-                        "hostUserId": { "$ref": "../definitions/userId.json" },
+                        "hostUserId": {
+                            "$ref": "../definitions/userId.json",
+                            "description": "which player will run the bot. It is the same as the player that added the bot."
+                        },
                         "allyTeam": { "type": "string" },
                         "team": { "type": "string" },
                         "player": { "type": "string" },


docs/schema/clan.md
```diff
@@ -0,0 +1,2338 @@
...
+
+# Clan
+
+Clans are a permanent kind of party with additional features.
+
+A clan has a profil which includes several information like tag, name, description. The following use-case diagrams gives an overview about the topics which has to be handled with this schema:
+
+![](assets/20251218_215155_Clan_Hub_Client_Server.svg)
+
+![](assets/20251218_215852_CLan_Hub_Sever_Client.svg)
```
geekingfrog: I am not a fan of including svg diagram there.
Personally I don't find them very useful, but the main issue is that they are effectively read only. If someone else wants to make any edit, they are kind of toast.

I wouldn't mind dropping them, but if you'd like to have some diagrams, I suggest you write them with [mermaid.js](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) which is supported by github (but not limited to github either).

docs/schema/clan.md
```diff
@@ -0,0 +1,2338 @@
...
+Every user is allowed to create a clan with [create](#create). Then he is the clan leader. The clan leader is the only one who is allowed to delete the clan with the request [remove](#remove).
+
+Clan leader and clan co-leaders are able to the profile [update](#update).
+
+Clan leader and clan co-leaders are allowed to invite different players with [invite](#invite). (There are no limitations of the count of members.) The invitation can be cancelled by [cancelInvite](#cancelinvite). Invited players can accept invitiation by [acceptInvite](#acceptinvite) and decline by [declineInvite](#declineinvite).
+
+Only the clan-leader is able to kick members by [kickMember](#kickmember). Similarly, any member can leave the clan they are currently in with [leave](#leave).
+
+The clan leader is allowed to appoint up to three members as clan co-leader with [setRole](#setrole). The clan leader has also the possibility to overtake his leadership to a different member with the same command. (ATTENTION: By executing this command the clan leader loses his permissions!)
+
```
geekingfrog: I think "has also the possibility to transfer his leadership role to" is more correct.

src/schema/clan/invite.ts
```diff
@@ -0,0 +1,23 @@
...
+            clanId: Type.Ref("clanId"),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "clan_not_available" },
+        { status: "failed", reason: "user_not_available" },
+        { status: "failed", reason: "no_permission" },
+        { status: "failed", reason: "already_invited" },
+        { status: "failed", reason: "already_member" },
```
geekingfrog: we already use `invalid_user` in a few places, which should be preferred over `user_not_available`. Similarly `invalid_clan` is more consistent.

I think `already_invited` should be dropped and the invite should succeed if there is already a pending invite. Because in this case, it's not like the user can do anything upon seeing the error, and it effectively means the action was carried out.

src/schema/clan/leave.ts
```diff
@@ -0,0 +1,21 @@
...
+    request: {
+        data: Type.Object({
+            userId: Type.Ref("userId"),
+            clanId: Type.Ref("clanId"),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "no_member" },
+        { status: "failed", reason: "permission_denied" },
```
geekingfrog: how can you be forbidden to leave the clan you're in?

src/schema/clan/remove.ts
```diff
@@ -0,0 +1,21 @@
...
+        data: Type.Object({
+            userId: Type.Ref("userId"),
+            clanId: Type.Ref("clanId"),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "no_member" },
+        { status: "failed", reason: "permission_denied" },
+        { status: "failed", reason: "clan_not_found" },
```
geekingfrog: `no_member` and `clan_not_found` are invalid reasons. There is always at least one member in a clan.

src/schema/clan/setRole.ts
```diff
@@ -0,0 +1,24 @@
...
+    request: {
+        data: Type.Object({
+            userId: Type.Ref("userId"),
+            clanId: Type.Ref("clanId"),
+            targetRole: Type.Ref("clanRole"),
+        }),
+    },
+    response: [
+        { status: "success" },
+        { status: "failed", reason: "only_one_leader" },
```
geekingfrog: `only_one_leader` cannot happen. Because if the leader set the leader role to someone else, they transactionally lose their leader role, and so the invariant of "only one leader" is held.

docs/schema/clan.md
```diff
@@ -0,0 +1,2338 @@
...
+Every user is allowed to create a clan with [create](#create). Then he is the clan leader. The clan leader is the only one who is allowed to delete the clan with the request [remove](#remove).
+
+Clan leader and clan co-leaders are able to the profile [update](#update).
+
+Clan leader and clan co-leaders are allowed to invite different players with [invite](#invite). (There are no limitations of the count of members.) The invitation can be cancelled by [cancelInvite](#cancelinvite). Invited players can accept invitiation by [acceptInvite](#acceptinvite) and decline by [declineInvite](#declineinvite).
+
+Only the clan-leader is able to kick members by [kickMember](#kickmember). Similarly, any member can leave the clan they are currently in with [leave](#leave).
+
+The clan leader is allowed to appoint up to three members as clan co-leader with [setRole](#setrole). The clan leader has also the possibility to overtake his leadership to a different member with the same command. (ATTENTION: By executing this command the clan leader loses his permissions!)
```
L-e-x-o-n: I think a limit on the number of clan members makes more sense than a limit on the number of roles. In any case the limit should be defined on the server, not the protocol.

src/schema/definitions/clan.ts
```diff
@@ -0,0 +1,20 @@
...
+        name: Type.String({ maxLength: 30 }),
+        description: Type.Optional(Type.String({ maxLength: 500 })),
+        logoUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        websiteUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        discordUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        teamspeakUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        twitchUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        youtubeUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        facebookUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
+        twitterUrl: Type.Optional(Type.String({ format: "uri", maxLength: 200 })),
```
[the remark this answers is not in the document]
L-e-x-o-n: Agreed. At least for the first version. eventually maybe it could be a list of URLs which could be more flexible and easier to format on the client.

docs/schema/clan.md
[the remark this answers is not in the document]
Rimek86: Oh ok. I was wondering about why i have to add my information after automatic creation of this file. This explains it. Thank's for the remark. I will add README.md to the folder path.

docs/schema/clan.md
Rimek86: Ok. Will remove the diagrams. I used it for me because i find the diagrams helpfull. But in that case i save the diagramms locally for me. Makes it easier for updating this file.

docs/schema/clan.md
Rimek86: Ok. I will remove this here. The limitation of co-leaders i will remove from my concept. But i think we should not implement a memmber limitation. Why you think we should do this? What is the use case?

src/schema/clan/changedRole.ts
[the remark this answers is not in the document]
Rimek86: I think you're right. Will remove this...

src/schema/clan/viewListChanged.ts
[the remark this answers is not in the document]
Rimek86: Good point. But then i would implement a "Update-Button". I would find it unpleasant if you had to switch to the new client to get an update.

src/schema/clan/leave.ts
```diff
@@ -0,0 +1,21 @@
...
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Leave your clan.",
+    request: {
+        data: Type.Object({
+            userId: Type.Ref("userId"),
+            clanId: Type.Ref("clanId"),
```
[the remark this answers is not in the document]
Rimek86: The reason to to this is cause i want to design the clan system without depentencies to the user profile. Means i don't want to save the clanid to the user profile. I want to save the userids in the clan. The benefit is that the profiles do not grow up with data. I think not all users wants to join a clan. So this data would be very often empty in the profile...

src/schema/clan/create.ts
```diff
@@ -0,0 +1,28 @@
...
+        { status: "failed", reason: "invalid_name_length" },
+        { status: "failed", reason: "invalid_description_length" },
+        { status: "failed", reason: "invalid_logo_url" },
+        { status: "failed", reason: "invalid_website_url" },
+        { status: "failed", reason: "invalid_discord_url" },
+        { status: "failed", reason: "invalid_teamspeak_url" },
+        { status: "failed", reason: "invalid_twitch_url" },
+        { status: "failed", reason: "invalid_youtube_url" },
+        { status: "failed", reason: "invalid_facebook_url" },
+        { status: "failed", reason: "invalid_twitter_url" },
```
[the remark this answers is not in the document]
Rimek86: VS Code show me an error that "details" is not available in FailedResponseSchema. 

But one thing has become clear to me. User input like this should be checked directly by the client and not by the server. The only error the server would need to report is that the clan tag or clan name is already taken. That's why I would like to use the details...

src/schema/clan/leave.ts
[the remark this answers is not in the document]
Rimek86: Wrote as reply of a comment of L-e-x-o-n:  I need a definition clanMember.ts i will add...

Rimek86:
I have rework my concept and replaces clanId with the clan object. Additionally i added clanMembers.ts to be able to append the role and membership to the user in an indepentend way (What do you mean. Is this way good or would it be better to directly integrate clanMemberRole and clanID to the user?)

Additionally i have a general question according to  a performance topic: If the client call clan/viewList i have all clan profiles sendet to the client and if the user want to see the details of the profil clan/view is not neccessary i think. Am i right?

src/schema/definitions/clan.ts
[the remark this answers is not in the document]
geekingfrog: That can work yes. It would require another request/response to get the list of available logo from server, but that's not too bad.

But again, I would drop the logo completely for now, it's already a very large change. Adding a logo later will be relatively easy.

src/schema/clan/leave.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Leave your clan.",
+    request: {
+        data: Type.Object({
+            user: Type.Ref("user"),
```
geekingfrog: shouldn't that be `userId`?

src/schema/clan/kick.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Kick a member from your clan.",
+    request: {
+        data: Type.Object({
+            user: Type.Ref("user"),
```
geekingfrog: `userId` is enough

src/schema/clan/deleted.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "A clan was deleted. Sent to all clan members.",
+    event: {
+        data: Type.Object({
+            clan: Type.Ref("clan"),
```
geekingfrog: `clanId`

src/schema/clan/delete.ts
```diff
@@ -0,0 +1,16 @@
...
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Delete your clan.",
+    request: {
+        data: Type.Object({
+            user: Type.Ref("user"),
+            clan: Type.Ref("clan"),
```
geekingfrog: neither `user` nor `clan` are required there.

src/schema/clan/cancel.ts
```diff
@@ -0,0 +1,16 @@
...
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Cancel an invite to your clan.",
+    request: {
+        data: Type.Object({
+            user: Type.Ref("user"),
+            clan: Type.Ref("clan"),
```
geekingfrog: `userId`, and `clan` isn't required.

src/schema/clan/README.md
```diff
@@ -0,0 +1,23 @@
+# Clan
+
+Clans are a permanent kind of party with additional features.
+
+A clan has a profil which includes several information like tag, name, description. The following use-case diagrams gives an overview about the topics which has to be handled with this schema:
```
geekingfrog: there is no more use-case diagrams.

docs/schema/clan.md
```diff
@@ -0,0 +1,3323 @@
...
+```
+</details>
+
+<details>
+<summary>Example</summary>
+
+```json
+{
+    "type": "request",
+    "messageId": "cillum Excepteur consectetur dolore commodo",
```
badosu: ?

src/schema/clan/viewList.ts
```diff
@@ -0,0 +1,18 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Get the list of clans",
```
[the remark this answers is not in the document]
L-e-x-o-n: No, I am pretty sure this is for getting the list of all clans. A player should only be a member of up to one clan at a time.

src/schema/definitions/clan.ts
```diff
@@ -0,0 +1,12 @@
+import { Type } from "@sinclair/typebox";
+
+export const clan = Type.Object(
+    {
+        clanId: Type.Ref("clanId"),
+        tag: Type.String({ minLength: 3, maxLength: 6 }),
+        name: Type.String({ maxLength: 30 }),
+        description: Type.Optional(Type.String({ maxLength: 500 })),
+        clanMembers: Type.Array(Type.Ref("clanMember")),
```
badosu: ditto for above, just `members`

src/schema/clan/view.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Get information about a clan.",
+    request: {
+        data: Type.Object({
+            clanId: Type.Ref("clanId"),
```
badosu: Is the pattern of passing `somethingId` on a request already scoped by `something` present in other tachyon requests?

docs/schema/clan.md
p2004a: this is auto generate via `json-schema-faker`.

src/schema/clan/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Manage clan
+
+Every user is allowed to create a clan with [clan/create](#create). Then he is the clan leader. The clan leader is the only one who is allowed to delete the clan with the request [clan/delete](#delete).
```
badosu: Why do we care about specifying behavior? Is this, and everything below this line, something _strictly_ required by the messaging contract or just an example of a possible implementation?

src/schema/clan/view.ts
badosu: > Is the pattern of passing `somethingId` on a request already scoped by `something`

In the diff above we can clearly see that we are passing a data structure containing a parameter `clanId` inside a request already scoped by `clan`, in this case: `clan/view`.

> present in other tachyon requests?

Is this a pattern already employed in other requests?

---

It seems the answer is no, as [exemplified in this request](https://github.com/beyond-all-reason/tachyon/blob/master/src/schema/lobby/join.ts#L12) although the pattern might be inconsistent.

Hence the answer to this question hopefully would be no and the recommendation to be to not use `clanId`, rather `id`.

I hope I was able to clarify. If not please let me know.

docs/schema/clan.md
badosu: Thank you.

src/schema/clan/view.ts
p2004a: We have both naming patterns already.

We should unify on one, I don't have preferences on which one.

badosu: That's great, thank you!

src/schema/clan/viewList.ts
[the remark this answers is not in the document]
geekingfrog: Correct, this should return all clans. And yes, there is no mechanism for pagination filter.
I think we should be fine for a while and then we can think of something. For example, there's something similar to get the entire list of lobbies. A back of the envelope calculation showed that even with hundreds of lobbies it's just not that much and overall should be manageable. I also did some load testing up to 800 lobbies and that was okay. So Clans, which should be more limited, should also be fine for a while.

src/schema/clan/README.md
```diff
@@ -0,0 +1,23 @@
+# Clan
+
+Clans are a permanent kind of party with additional features.
```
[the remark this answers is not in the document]
geekingfrog: See the [readme for parties](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/party.md).

src/schema/definitions/clan.ts
```diff
@@ -0,0 +1,12 @@
+import { Type } from "@sinclair/typebox";
+
+export const clan = Type.Object(
+    {
+        clanId: Type.Ref("clanId"),
```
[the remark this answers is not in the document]
Rimek86: But in user.ts we also use userId? (userId: Type.Ref("userId"),)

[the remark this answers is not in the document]
Rimek86: So i prefere to stay at this concept... @geekingfrog

src/schema/clan/create.ts
```diff
@@ -0,0 +1,13 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Create a clan.",
+    request: {
+        data: Type.Object({ clanBaseData: Type.Ref("clanBaseData") }),
```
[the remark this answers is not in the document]
Rimek86: The client can set the id to 0 and server ignored it. So i think it is not realy an issue. I think it would be excessive to create special structures for this one use case.

src/schema/clan/leave.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Leave your clan.",
+    request: {
+        data: Type.Object({
+            userId: Type.Ref("userId"),
```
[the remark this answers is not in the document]
Rimek86: I see. Will be changed.

src/schema/clan/updated.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "A clan was updated. Sent to all clan members.",
+    event: {
+        data: Type.Object({
+            changeDescription: Type.String(),
```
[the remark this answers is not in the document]
Rimek86: changeDescription should be a free string field containing a custom message for the event to ensure maximum flexibility and has nothing to do with clan description. But Lexon wrote to me on Discord and I'm going to do the whole thing differently and create specific events instead of a generic one...

src/schema/definitions/clanRole.ts
```diff
@@ -0,0 +1,10 @@
+import { Type } from "@sinclair/typebox";
+
+export const clanRole = Type.Enum(
+    {
+        Member: "member",
+        CoLeader: "coLeader",
+        Leader: "leader",
+    },
```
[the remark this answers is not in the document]
Rimek86: Enum is more strict. UnionEnum creats a list of possible values.

src/schema/clan/create.ts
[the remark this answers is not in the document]
p2004a: That's not great, use `Type.Omit`.

src/schema/clan/update.ts
```diff
@@ -0,0 +1,13 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Update your clan.",
+    request: {
+        data: Type.Object({ clan: Type.Ref("clan") }),
```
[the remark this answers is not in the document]
geekingfrog: I think we want the protocol to be as clear as possible. Reusing slightly different structure where the meaning and/or value of field depend on which context they are used is not great. It's confusing and error prone.
I'd rather have specific objects/structs for specific purpose, even if it feels a bit verbose.

src/schema/clan/kicked.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "A player was kicked from a clan. Sent to the kicked player.",
+    event: {
+        data: Type.Object({
+            clanBaseData: Type.Ref("clanBaseData"),
```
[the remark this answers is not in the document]
Rimek86: I thought about it. Do i realy need the id? Because the user profil has the information of clan membership i think. So the client nows about the name of the clan and the event triggers a message in the client and removes the entries of the profil page in the client. Or do we need id for verification topics on client side?

src/schema/definitions/clanRole.ts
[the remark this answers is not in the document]
Rimek86: @p2004a I'm sorry. You are right. I made the mistake and trust Copilot. I think this article get a good overview about the differences (https://medium.com/@sanyagubrani/typescript-enums-or-union-types-which-one-should-you-use-377ecae947af) or? If i understand this correctly UnionEnum is the better choice for us correct?

.vscode/settings.json
```diff
@@ -1,11 +0,0 @@
...
-    "editor.formatOnSave": true,
-    "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": "explicit"
-    },
-    "eslint.validate": ["typescript"],
-    "[typescript]": {
-        "editor.defaultFormatter": "esbenp.prettier-vscode"
-    },
-    "typescript.tsdk": "node_modules\\typescript\\lib"
-}
```
geekingfrog: There is a `settings.json` file in the repo. I am not using vscode so I'm not sure how configurable/usable it is to leave that kind of config in the repo.
Not against removing it, though I think it would be clearer to leave that to a different PR and restore this file on this branch.

src/generate-js-validators.ts
```diff
@@ -110,6 +111,7 @@ function ucs2length(str) {
             esm: false,
         },
         keywords: ["tachyon"],
+        strict: false, // Disable strict mode to allow unknown keywords
```
geekingfrog: What is that? Why do you need this?

src/generate-json-schemas.ts
```diff
@@ -277,7 +277,14 @@ export async function generateJsonSchemas(): Promise<TachyonConfig> {
...
         // We need to have ternary in mapper below because the schema might have been modified by us already.
-        mapRefs(schema, (ref) => `../definitions/${ref}.json`);
+        mapRefs(schema, (ref) => {
+            // but the ref may have already been "processed". This can happen when a definition
+            // embeds another definition
+            if (ref.endsWith(".json")) {
+                return ref.replace(/.*\/definitions\/(.*)\.json/, "../definitions/$1.json");
+            }
+            return `../definitions/${ref}.json`;
+        });
```
geekingfrog: Is this coming from https://github.com/beyond-all-reason/tachyon/pull/88 ? And is it required for the clan schemas?
Because it seems this implementation isn't quite correct. I'll try to have another look in the coming days and properly address the problem then.

src/schema/clan/kicked.ts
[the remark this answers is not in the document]
geekingfrog: So, purely theoretically you could have some *mad* race condition where:

* kick player from clan A
* `clan/kicked` stuck server side somehow
* player creates/join another clan B
* finally player receives `clan/kicked`

without the clan ID, the player would think they've been kicked from clan B, which is not the case.
Admittedly this is rather far fetch, but sending the clan ID doesn't cost much and remove that kind of uncertainty, it's just generally nice to have that.

src/schema/clan/leave.ts
```diff
@@ -0,0 +1,13 @@
+import { Type } from "@sinclair/typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Leave your clan.",
+    request: {
+        data: Type.Object({}),
```
geekingfrog: I believe you can just do `data: {}`, no need for the empty object.

src/schema/clan/update.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Update your clan.",
+    request: {
+        data: Type.Object({
+            clan: Type.Omit(Type.Ref("clan"), ["clanId", "membersCount", "members"]),
```
geekingfrog: It's going to end up the same in the json schema, but I prefer having the stuff you can update listed explicitly instead of inferring them through omission of other stuff.
Later, if we add some other property to clan, I want the ability to update that property be a deliberate choice, and not a side effect of this construction. For example, if we add `createdAt`, it's be very easy to miss and forget to remove it there.

docs/schema/clan.md
```diff
@@ -2049,6 +2049,7 @@ export interface ClanUpdateableBaseData {
 export interface ClanMember {
     userId: UserId;
     role: ClanRole;
+    joinDate: string;
```
geekingfrog: We already have a [unixTime](https://github.com/geekingfrog/tachyon/blob/0985e3a3e855a93b299313db615380f380b51e51/src/schema/definitions/unixTime.ts) to represent time. And in this case `joinedAt` is probably better. The client can then decide to only show a date.

src/schema/clan/update.ts
```diff
@@ -7,9 +7,7 @@ export default defineEndpoint({
     target: "server",
     description: "Update your clan.",
     request: {
-        data: Type.Object({
-            clanUpdateableData: Type.Ref("clanUpdateableData"),
-        }),
+        data: Type.Intersect([Type.Ref("clanUpdateableData")]),
```
geekingfrog: Same comment, Intersect with only one ref is useless and generate confusing json schema.

src/schema/clan/create.ts
```diff
@@ -7,9 +7,7 @@ export default defineEndpoint({
     target: "server",
     description: "Create a clan.",
     request: {
-        data: Type.Object({
-            clanUpdateableBaseData: Type.Ref("clanUpdateableBaseData"),
-        }),
+        data: Type.Intersect([Type.Ref("clanUpdateableBaseData")]),
```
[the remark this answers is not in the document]
Rimek86: My idea was that on creation time only the mandatory fields has to be entered. But i think if someone creates a new clan he can leave the description field emtpy. So i'm fine with...

PR #88, opened 2026-01-05
Fix schema generation with embedded definitions

geekingfrog:
When a definition has a ref to another definition, it would produce an invalid path like
`../definitions/../../definitions/clanBaseData.json.json` so detect this case and adjust the logic to get the correct path.

I do not fully understand the whole logic behind all the `mapRefs`, so there may be a better way to achieve the same result, I'm not sure how.

### DIFF
--- a/schema/system/disconnect/request.json
+++ b/schema/system/disconnect/request.json
@@ -19,5 +19,5 @@
             "required": ["reason"]
         }
     },
-    "required": ["type", "messageId", "commandId"]
+    "required": ["type", "messageId", "commandId", "data"]
 }
--- a/docs/schema/autohost.md
+++ b/docs/schema/autohost.md
@@ -76,6 +76,8 @@ Request to add a new player to the battle.
 
 ```json
 {
+    "$id": "https://schema.beyondallreason.dev/tachyon/autohost/addPlayer/request.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "AutohostAddPlayerRequest",
     "tachyon": {
         "source": "server",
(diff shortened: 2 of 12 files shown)


Rimek86 (approved):
I have tested your changes with my code and it seems to work well.

p2004a (changes requested):
The problem is because we are dynamically doing `import` on source, in js when you do an import of the same library, it's the same underlying object. `mapRefs` doesn't clone, but just overrides in place.

The `user` definition is loaded in `src/schema/user/updated.ts`, processed there, and then again when directly loading the `src/schema/definitions/user.ts`.

So, to solve this, we have to either make replacements idempotent or we need to check if something like [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) on schema object would work.

I would approve both, the approach in this PR is the independence approach, but it should be extended to all the other locations of `mapRefs` as the same problem will happen if two requests import some other definitions. Maybe just like

```ts
mapRefs(schema, (ref) => ref.endsWith('.json') ? ref : `../definitions/${ref}.json`);
```

geekingfrog:
Alright, I reworked the commit to instead perform deep clone on all imported definitions.
I checked by adding another ref from a definition and it compiled correctly. This should unblock https://github.com/beyond-all-reason/tachyon/pull/85

schema/system/disconnect/request.json
```diff
@@ -19,5 +19,5 @@
             "required": ["reason"]
         }
     },
-    "required": ["type", "messageId", "commandId"]
```
geekingfrog: uuuh, why is `data` now a required property there???

p2004a: eh...

I will fix it.

Issue #100, opened 2026-02-22
Player status is missing upon lobby/join


Hectate:
PR #84 added player status fields for ``lobby/updated`` events, however the response received for ``lobby/join`` does not include this information. I realize these are optional fields, but the problem is now that a late-joining lobby member will not get any information about this until a ``lobby/updated`` event is received.
I believe that, should this information be available, we would want to also send it upon ``lobby/join``. An argument could be made either way for ``lobby/create`` responses, since they're largely the same as ``lobby/join`` except that it's known to be a completely fresh lobby containing only one player (assuming we don't later add something for creating and joining a lobby as a party automatically).

Comparison:
```ts
    players: {
        [k: string]: {
            id: UserId;
            allyTeam: string;
            team: string;
            player: string;
        };
    };
```

```ts
    players?: {
        [k: string]: {
            id: UserId;
            allyTeam?: string;
            team?: string;
            player?: string;
            isReady?: boolean;
            assetStatus?: "missing" | "downloading" | "ready";
        } | null;
    };

PR #101, opened 2026-02-26
Add player status to lobby details

geekingfrog:
Fix issue #100

### DIFF
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -610,7 +610,7 @@ export interface StartBox {
                 "team": "ut esse ea qui Lorem",
                 "player": "qui eiusmod",
                 "isReady": true,
-                "assetStatus": "ready"
+                "assetStatus": "complete"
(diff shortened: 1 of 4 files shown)


Hectate:
The PR looks good to me.

I was caught off guard briefly because we have both ``assetStatus: "ready"`` and ``isReady: true`` and I thought they were redundant - until I remembered that "Player is ready" is different than "Player has all the assets".

To avoid any confusion, could we add to this PR that the ``assetStatus`` values be changed in all places to ``["missing", "downloading", "complete"]`` instead?

Hectate (approved):
Thanks. Looks good to me.

Hectate:
We also have a line in the docs to update: https://github.com/beyond-all-reason/tachyon/blob/33ed6d4bfed63137231fe35e3c0c6bdf969bbc8a/docs/schema/lobby.md?plain=1#L105

Issue #87, opened 2026-01-03
Figure out modoptions for lobbies


geekingfrog:
In order to support nutty raptor, scavenger, and whatever lobby customisation, we need to support modoptions is lobbies.

PR #102, opened 2026-02-28
Support for mod options. Address #87

geekingfrog:
https://github.com/beyond-all-reason/tachyon/issues/87


geekingfrog:
After looking a bit more into how this system works, it turns out the game itself is in full control of the options and how they should be displayed.
So keep the protocol to a minimum and let the client use the game data to drive the UI.

Hectate (approved):
I see no issues from here.

PR #109, opened 2026-03-14
Add a request to add specs to the currently running game

geekingfrog:
So one can join a running game even if they weren't part of the lobby at start.

### DIFF
--- a/src/schema/lobby/joinBattle.ts
+++ b/src/schema/lobby/joinBattle.ts
@@ -1,0 +1,15 @@
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description:
+        "Join the battle this lobby is currently playing (as a spectator). If the request success, the client should receive shortly after a battle/start request from the server with the data required to join the game.",
+    request: {},
+    response: [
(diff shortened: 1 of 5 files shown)


src/schema/lobby/joinBattle.ts
```diff
@@ -0,0 +1,18 @@
+import Type from "typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "user",
+    target: "server",
+    description: "Join the game as a spectator this lobby is currently playing.",
+    request: {},
```
geekingfrog: later down the road we can add some data there if one wants to join the battle as a player, either to replace someone who left or for some archon/special setup.
But this also require some checks about permissions as these commands can be abused, so for now just ignore the whole thing while leaving the ability to address this later.

Hectate:
I'm looking at the ``LobbyJoinBattleOkResponseData `` and it occurs to me that this might be better treated the way that we do ``lobby/startBattle`` and ``battle/start``. In fact, wouldn't it be possible for the user to just get a ``battle/start`` event (or maybe ``battle/join``?) if they send a successful ``lobby/joinBattle`` request?

PR #120, opened 2026-03-22
Added ``maxMembers`` to party states

Hectate:
The server-side setting of maximum members per party was not exposed to the client, making them unaware that invites over this limit would consistently fail. This exposes the ``maxMember`` value to the client.
~~I also removed a spurious ``not_in_party`` failure response for ``cancelInvite``. It should not be possible to fail to cancel an invite because an invitee is "not in the party".~~ Error message was kept for if the requestor was not in a party.

### DIFF
--- a/src/schema/party/README.md
+++ b/src/schema/party/README.md
@@ -6,6 +6,8 @@ together and during lobby balancing.
 A player can only be in at most one party at any given time.
 When the last member of a party leaves, the party is disbanded.
 
+Parties have a maximum player count of ``maxMembers``, defined by the Tachyon server. All current members and pending invitations count towards the limit. For example, a party of 2 members with 1 pending invite will not successfully send new invites if ``maxMembers`` is 3.
+
 Any change to the party member or the pending invites is propagated with a [party/updated](#updated) event sent
 to all connected members and invited players.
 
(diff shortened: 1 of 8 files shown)


src/schema/party/cancelInvite.ts
```diff
@@ -11,9 +11,5 @@ export default defineEndpoint({
             userId: Type.Ref("userId"),
         }),
     },
-    response: [
-        { status: "success" },
-        { status: "failed", reason: "not_in_party" },
-        { status: "failed", reason: "invalid_invite" },
-    ],
+    response: [{ status: "success" }, { status: "failed", reason: "invalid_invite" }],
```
geekingfrog: you still need the `not_in_party` error in case the client sends a `party/cancelInvite` but is not itself in a party.

src/schema/party/README.md
```diff
@@ -6,6 +6,8 @@ together and during lobby balancing.
 A player can only be in at most one party at any given time.
 When the last member of a party leaves, the party is disbanded.
 
+Parties have a maximum player count of ``maxMembers``, defined by the Tachyon server. All current members and pending invitations count towards the limit. For example, a party of 2 members with 1 pending invite will not successfully send new invites if ``maxMembers`` is 3.
```
geekingfrog: why the double backticks there? Isn't 1 enough?

geekingfrog (approved):

PR #125, opened 2026-03-25
Fix for issue 105

### LINKED ISSUE #105: Small updates needed to Clan documentation
@Rimek86 I thought drop some feedback on things I noticed.

1. ``A clan has a profil which includes...`` << "profile" is misspelled.
2. ``Invited players can accept invitiation by [aclan/cceptInvite](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/clan.md#acceptinvite)`` << ``aclan/cceptInvite`` the 'a' is moved out of the correct place.
3. ``Only the clan-leader is able to kick members`` << This explicitly excludes co-leaders. Is that correct?
4. What is expected to happen if ``clan/kickMember`` is used on one's self? Is that an ``invalid_request`` failure?
5. Is ``clan/setRole`` also used to demote a ``coLeader`` back to ``member``? We should explicitly say if so or not.
6. I dislike that ``clan/setRole`` does double-duty as a "promote" and "promote and also vacate self as leader" request. We could separate these actions in the client for safety, but I feel like having a separate ``setLeaderRole`` request that explicitly will also demote yourself is better. Just a thought.
7. After a self-demotion from ``leader`` role, what is the former leader's role? Are they a ``coLeader`` or just a ``member``? We should explicitly say which.

Rimek86:
PR for https://github.com/beyond-all-reason/tachyon/issues/105 

- fail reasons added to clan/accept, clan/create and clan/update. (The first change is from issue 105. The other ones i identified when i create the create form in new client. Because a dedicated failed reason if a user tries to create a new clan with invalid values is helpfull.)
- clan/viewInvites added as new command based on the feedback of issue 105


Rimek86 (commented):
> Sorry for the repeated review comments, I realized afterward that I should have started with files in the ``src/`` folders and not reviewed the generated files 🤦  

No Problem. I set this PR now to "Ready to review" because i lost the overview about all my open PRs. So i try to reduce it now...

src/schema/clan/create.ts
```diff
@@ -9,5 +9,11 @@ export default defineEndpoint({
     request: {
         data: Type.Ref("clanUpdateableData"),
     },
-    response: [{ status: "success" }],
+    response: [
+        { status: "success" },
+        {
+            status: "failed",
+            reason: "incorrect_data",
```
Hectate, line 16: Why are we using ``incorrect_data`` here as a new reason when we already have ``invalid_request``?

Hectate (changes requested):
Looks fairly good excluding the spurious failure reason.

Issue #112, opened 2026-03-19
[placeholder] Boss for lobbies


geekingfrog:
This is a **placeholder** issue. We are pretty sure we want this feature, but need to figure how the exact how.

Hectate:
My initial thought is that we need to add an optional ``bossUserId: string`` property to the lobby where the value is the ID of the client holding the boss role. The client would simply check if their ID matches to enable boss features. The server would reject boss requests from non-boss clients with ``unauthorized``.

The alternative would be to add it to the ``players`` and ``spectators`` as an optional ``isBoss: boolean`` property for the appropriate user, but I prefer having it separated. If it's mixed in with the other clients information directly, then we'd have to constantly check all clients in those lists for the boolean on every update - in case the boss was added, removed, or changed. Having the ID is a lot more direct and immediate.

And to answer the possible question, ``spectators`` should be capable of being a boss. Event/Tournament organizers will need to be able to manage their lobbies as a boss without being part of the ``players``.

PR #126, opened 2026-04-05
Support for boss in lobbies

geekingfrog:
issue: https://github.com/beyond-all-reason/tachyon/issues/112

### DIFF
--- a/src/schema/definitions/lobbyDetails.ts
+++ b/src/schema/definitions/lobbyDetails.ts
@@ -33,6 +33,8 @@ export const lobbyDetails = Type.Object(
                 }
             )
         ),
+        areBossesEnabled: Type.Boolean(),
+        bosses: Type.Record(Type.String(), Type.Object({})),
         players: Type.Record(
             Type.String(), // userId, using Type.Ref() generates a schema with only not: {}
             Type.Object({
--- a/src/schema/lobby/README.md
+++ b/src/schema/lobby/README.md
@@ -101,9 +101,17 @@ through `joinAllyTeam` or with the join queue), the server will automatically as
 `{"isReady": false, "assetStatus": "complete"}`. If this is incorrect the client should send a request to correct it.
 
 
+### Lobby bosses
+
+Lobbies can have "bosses", so that a player, or a group of friends, can setup a game just like they want
+with full control. When one or more boss are present in a lobby, some commands may be disabled for non-boss players.
+The creator of a lobby can choose to forbid bosses in the lobby to ensure all changes go through votes.
+This setting cannot be changed after creation.
+
+
 ### Lobby updates
 
-Any member can update most (any?) property of the lobby. Updates are all or nothing, if updating a proprety
+Any member can update most (any?) property of the lobby unless bosses are present. Updates are all or nothing, if updating a proprety
 is not possible (invalid or forbidden), then no update take place.
 The result of the updates is then transmitted to all members via [lobby/updated](#updated) events.
 
--- a/src/schema/lobby/create.ts
+++ b/src/schema/lobby/create.ts
@@ -11,6 +11,7 @@ export default defineEndpoint({
             name: Type.String(),
             mapName: Type.String(),
             allyTeamConfig: Type.Ref("allyTeamConfig"),
+            areBossesEnabled: Type.Boolean({ default: true }),
         }),
     },
     response: [
--- a/src/schema/lobby/updated.ts
+++ b/src/schema/lobby/updated.ts
@@ -61,6 +61,7 @@ export default defineEndpoint({
                     )
                 )
             ),
+            bosses: Type.Optional(Type.Record(Type.String(), Nullable(Type.Object({})))),
             players: Type.Optional(
                 Type.Record(
                     // this is a userId, but using Type.Ref("userId") leads to a schema with only: `not: {}`
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -105,9 +105,17 @@ through `joinAllyTeam` or with the join queue), the server will automatically as
 `{"isReady": false, "assetStatus": "complete"}`. If this is incorrect the client should send a request to correct it.
 
 
+### Lobby bosses
+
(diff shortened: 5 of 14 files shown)


src/schema/lobby/updated.ts
```diff
@@ -61,6 +61,7 @@ export default defineEndpoint({
                     )
                 )
             ),
+            bosses: Type.Optional(Type.Record(Type.String(), Nullable(Type.Object({})))),
```
geekingfrog, line 64: the lack of `areBossesAllowed` is by design there since this flag is set in stone at lobby creation.

src/schema/definitions/lobbyDetails.ts
```diff
@@ -33,6 +33,8 @@ export const lobbyDetails = Type.Object(
                 }
             )
         ),
+        areBossesAllowed: Type.Boolean(),
+        bosses: Type.Record(Type.String(), Type.Object({})),
```
geekingfrog, line 37: little bit awkward to have an empty object as the value, but I don't see how else we can represent a set.

src/schema/lobby/README.md
```diff
@@ -101,6 +101,14 @@ through `joinAllyTeam` or with the join queue), the server will automatically as
 `{"isReady": false, "assetStatus": "complete"}`. If this is incorrect the client should send a request to correct it.
 
 
+### Lobby bosses
+
+Lobbies can have "bosses", so that a player, or a group of friends, can setup a game just like they want
+with full control. When one or more boss are present in a lobby, some commands may be disabled for non-boss players.
+The creator of a lobby can choose to forbid bosses in the lobby to ensure all changes go through votes.
+This setting cannot be changed after creation.
```
Hectate, line 109: Do bosses vote between themselves or are all bosses able to directly issue a change regardless of how many other bosses may be present?

src/schema/lobby/updated.ts
Hectate: So ``bosses`` will simply be empty in non-boss lobbies?

src/schema/lobby/README.md
```diff
@@ -101,6 +101,14 @@ through `joinAllyTeam` or with the join queue), the server will automatically as
...
+
+
 ### Lobby updates
 
 Any member can update most (any?) property of the lobby. Updates are all or nothing, if updating a proprety
```
Hectate: "Any member can update most (any?) property of the lobby, unless bosses are enabled."

Hectate (dismissed):
A couple more points that I didn't see addressed anywhere:
* What happens if the only member of ``bosses`` leaves the lobby?
* We need to change ``lobby/listUpdated`` to indicate if a lobby is in boss mode or not.

src/schema/lobby/README.md
geekingfrog: I think this should be left to implementor decision. Codifying all boss behaviour at the protocol level is going to be too rigid. Just the discussion we had on discord showed there were many different viewpoints about what ppl want/need. I'd like to be able to change behaviour without having to change the protocol when possible.

geekingfrog:
>     * What happens if the only member of `bosses` leaves the lobby?
nothing special. Anyone can appoint a new boss. The details are left to the implementations I'd say.


>     * We need to change `lobby/listUpdated` to indicate if a lobby is in boss mode or not.
What do we want exactly there? The `areBossesAllowed` flag? The presence of an actual boss in the lobby?

src/schema/lobby/README.md
Hectate: I don’t disagree but the client will need to know who should see a voting option so it’s still something to think about.

Hectate:
> What do we want exactly there? The `areBossesAllowed` flag? The presence of an actual boss in the lobby?

The flag will be enough. I’m not interested in having to get updates if bosses come or go, but only if it’s enabled since it can’t be changed after.

src/schema/definitions/lobbyDetails.ts
Hectate: We might as well use an Array? I don't think that it'll be updated too frequently nor be very populated most of the time.

src/schema/lobby/create.ts
```diff
@@ -11,6 +11,7 @@ export default defineEndpoint({
             name: Type.String(),
             mapName: Type.String(),
             allyTeamConfig: Type.Ref("allyTeamConfig"),
+            areBossesAllowed: Type.Boolean({ default: true }),
```
Hectate: Could we make this ``areBossesEnabled`` instead?

src/schema/lobby/README.md
geekingfrog: In the lobby details (and updates), the `currentVote` has [the list of voters](https://github.com/geekingfrog/tachyon/blob/97285cd18587f99b278538cc09f5c3385b1e730c/src/schema/definitions/lobbyDetails.ts#L91-L98).
Unless I don't understand your question?

src/schema/definitions/lobbyDetails.ts
geekingfrog: That's valid. Though we use the merge patch approach everywhere else, so at least this is consistent.
I don't really know tbh, I think both approaches can be valid.
@p2004a do you have any insight on that?

p2004a: You could make it a boolean property on player/spectator instead

if not that, then both object and array would be fine:
- this array is short, so won't be a problem to just always replace
- using object, you can later add some additional properties to the object if desired, e.g. if boss stops being yes/no (I don't have use case, just hypothesizing)

src/schema/lobby/README.md
Hectate: I actually forgot about the list of voters. I’m good with that. So yeah when the client checks who can vote it won’t matter if boss mode is enabled or not.

Hectate (approved):
I'm good with how this is now.

src/schema/definitions/lobbyDetails.ts
geekingfrog: I'll leave it as that. We get the consistency of merge patch, and the empty object, as you mentioned, may be useful in the future.

PR #131, opened 2026-04-18
Add tags to lobbies

geekingfrog:
Keep the actual list of tags outside the protocol. We can hardcode the supported tags at first, and later if needed, have a way for clients to discover supported tags, though that may be very overkill.

### DIFF
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -543,6 +543,12 @@ Create a lobby
                             "required": ["value"]
                         }
                     }
+                },
+                "tags": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^.*$": { "type": "object", "properties": {} }
+                    }
                 }
             },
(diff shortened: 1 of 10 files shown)


Hectate:
Agreed that the protocol should not need to define tags. As you say, we need discoverability somehow, or else simply allow the client's to send in ad-hoc tags upon creation, but that's for a different discussion.

Hectate:
Just a couple of thoughts:
* Shouldn't we have the ``tags`` Record in ``lobby/update`` too?
* I presume the intent is to have an empty ``tags: {}`` if there are none, rather than making it optional for ``lobby/create`` (and ``lobby/update``)?

geekingfrog:
Good points.
I have made the tags optional at creation, the server will default to `{}` if nothing is provided.
And I've added the tags as json-merge-patch for `lobby/update`, I did forget that indeed.

Hectate (approved):

PR #135, opened 2026-04-23
Kick (and ban) request players from lobby

geekingfrog:
The overview:

* same request to kick and ban, the difference is marked with an optional field `banUntil`
* lobbies have a list of banned player. Not sure about that, that may be useful, but also could be dropped.
* one of the limitation of our schemas is that we can't put additional data in the error results. So when one joins a lobby but is banned, we can only say "you are banned", but not until when.

### DIFF
--- a/src/schema/lobby/left.ts
+++ b/src/schema/lobby/left.ts
@@ -9,8 +9,14 @@ export default defineEndpoint({
         "Sent by the server when the player is removed from the lobby. Can be kicked/ban or the lobby crashed",
     event: {
         data: Type.Object({
-            id: Type.String(),
+            id: Type.String({ description: "lobby id" }),
             reason: Type.String(),
+            bannedUntil: Type.Optional(
+                Type.Ref("unixTime", {
+                    description:
+                        "if this field is set, it means the client has been banned from this lobby until that timestamp.",
+                })
+            ),
         }),
     },
 });
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -140,6 +140,7 @@ In practice, this event should rarely be seen.
 - [joinAllyTeam](#joinallyteam)
 - [joinBattle](#joinbattle)
 - [joinQueue](#joinqueue)
+- [kickban](#kickban)
 - [leave](#leave)
 - [left](#left)
 - [listReset](#listreset)
@@ -985,6 +986,11 @@ export type VoteActions =
     | {
           type: "appointBoss";
           bossId: UserId;
+      }
+    | {
+          type: "kickban";
+          userId: UserId;
+          banUntil?: UnixTime;
       };
 export type VoteOutcomes = "passed" | "failed" | "cancelled" | "timeout";
 
@@ -1193,6 +1199,7 @@ export interface LobbyJoinRequestData {
                 "reason": {
                     "enum": [
                         "lobby_full",
+                        "banned",
                         "internal_error",
                         "unauthorized",
                         "invalid_request",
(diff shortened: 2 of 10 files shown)


L-e-x-o-n:
I would switch things up a bit.

A kick usually just removes you from the lobby but allows immediate rejoining, a ban doesn't allow rejoining for a set time if at all, therefore I would change the command from kick to ban.
Leaving `banUntil` empty would then mean an indefinite ban, a kick can be performed by putting a very short `banUntil`.
IMO ban list is unnecessary, we will likely want global lobby ban and then this list could get quite long.

geekingfrog:
What do you mean by global lobby ban?

L-e-x-o-n:
> What do you mean by global lobby ban?

Moderators currently have the ability to ban users from joining any lobby. This is currently used instead of a login ban (when it isn't a permanent one) because it allows users to login and acknowledge warning + check their restrictions.
It's up to debate whether the same approach will be wanted and necessary with tachyon but having the option seems useful.

geekingfrog:
Ok, I reworked the thing, it's a ban with a non optional `until`, put that in the past for only kicking.
I removed the list of banned player. Though in my mind it was only for players specifically banned through the action, and not publish all moderation actions.

Hectate:
> > What do you mean by global lobby ban?
> 
> Moderators currently have the ability to ban users from joining any lobby. This is currently used instead of a login ban (when it isn't a permanent one) because it allows users to login and acknowledge warning + check their restrictions. It's up to debate whether the same approach will be wanted and necessary with tachyon but having the option seems useful.

Since this is a lobby specific endpoint, I think the idea of a global ban shouldn't enter here. What we would do for that is probably add some properties to ``user/self`` responses that indicate to the client that there are restrictions on their online presence, which can be extended to all forms of moderation we end up implementing.

src/schema/lobby/left.ts
```diff
@@ -9,8 +9,14 @@ export default defineEndpoint({
         "Sent by the server when the player is removed from the lobby. Can be kicked/ban or the lobby crashed",
     event: {
         data: Type.Object({
-            id: Type.String(),
+            id: Type.String({ description: "lobby id" }),
             reason: Type.String(),
```
Hectate, line 13: Since this is getting edited, it's worth asking here. Do you feel that we should have a specific list of reasons rather than an overall string? We won't be able to i18n a raw string. I'd almost feel like this should be explicitly ``kickban``, ``crash``, ``server_shutdown``, ``lobby_deleted``, etc...

Edit: discussed in Discord. Not going to address this at this time.

L-e-x-o-n:
> > > What do you mean by global lobby ban?
> > 
> > 
> > Moderators currently have the ability to ban users from joining any lobby. This is currently used instead of a login ban (when it isn't a permanent one) because it allows users to login and acknowledge warning + check their restrictions. It's up to debate whether the same approach will be wanted and necessary with tachyon but having the option seems useful.
> 
> Since this is a lobby specific endpoint, I think the idea of a global ban shouldn't enter here. What we would do for that is probably add some properties to `user/self` responses that indicate to the client that there are restrictions on their online presence, which can be extended to all forms of moderation we end up implementing.

I didn't mean to suggest bringing global lobby bans here, just that there are potential sources of lobby bans *outside* of the in lobby command and in that case having a list of banned users *here* may be inconvenient/incorrect/unnecessary.
I agree with the rest.

geekingfrog:
I've made the `banUntil` field optional to merely mark a kick as different from a ban. Not a fan of magic values to carry semantics, though if a client set a time in the past, it would behave the same as a kick.

Hectate (approved):

PR #138, opened 2026-05-05
added request `user/report`

### LINKED ISSUE #137: Moderation: User Reporting MVP
Mod team has asked for moderation features in the new client. After discussion, it was agreed that an initial MVP deliverable would be "right click, select Report User, and submit a simple form".

This issue is to identify that the protocol will need to have a way to receive moderation activity from the client. That request will need:

1. ID(s) to report
2. Some overarching "reason"
3. Additional ad-hoc string of text from the user

Upon receipt of the report, the moderation team expects to be notified via Discord and be able to view the report details on the server webpage for moderation.

Hectate:
Initial request with simple content, and two basic types.

Closes #137

### DIFF
--- a/src/schema/definitions/moderationType.ts
+++ b/src/schema/definitions/moderationType.ts
@@ -1,3 +1,0 @@
-import { Enum } from "typebox";
-
-export const moderationType = Enum(["chat", "action"], { $id: "moderationType" });
--- a/src/schema/definitions/moderationType.ts
+++ b/src/schema/definitions/moderationType.ts
@@ -1,0 +1,3 @@
+import { Enum } from "typebox";
+
+export const moderationType = Enum(["chat", "action"], { $id: "moderationType" });
--- a/docs/schema/user.md
+++ b/docs/schema/user.md
@@ -3,6 +3,7 @@
 # User
 
 - [info](#info)
+- [report](#report)
 - [self](#self)
 - [subscribeUpdates](#subscribeupdates)
 - [unsubscribeUpdates](#unsubscribeupdates)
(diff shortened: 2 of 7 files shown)


src/schema/definitions/moderationType.ts
```diff
@@ -0,0 +1,3 @@
+import { Enum } from "typebox";
+
+export const moderationType = Enum(["chat", "action"], { $id: "moderationType" });
```
geekingfrog: I am actually not sure it's wise to put the actions hardcoded into the protocol. I believe a string should be enough, and then it's left to implementations to define what are the correct actions.
For example, I can easily imagine extending these to other things, like "early leaving/dodging", or "lobby action" vs "game action" or whatever.

Perhaps a better compromised would be to have `moderationReason: Object({type: string})`. This would allow us to change it to a `Union` if we need to pass additional data for some action, for example, the timestamp of the action.

Hectate: This setup was essentially a minimum viable product version of the team’s requested end product. The types enumerated here are relevant because they categorize the reports as one of two types:
1. “Chat” reports are literally “someone said something in a chat that was inappropriate 
2. “Action” reports are “someone did an action in game that was inappropriate”

Depending on which it is, the mods would either need to review those chat sources or they would need to watch a replay. Having that known up front would help with the categorization on the incoming reports and allow different mods to respond depending on the type of issue.

Hectate: It’s worth noting that there will definitely be more data needed in the future, but I didn’t want to pack too many things into the protocol before we would be able to effectively deliver them in the client and server.
There’s more details regarding scope in Discord, of course.

geekingfrog: You can still have them know upfront without the values set in the protocol. We don't need to make it fully backend driven and customisable.
We can have an agreement that the only supported actions are `chat` and `action`, and both teiserver and the client have these hardcoded, while the protocol doesn't enforce these.
That's essentially my point, keep the protocol extensible and loose regarding the exact details of the data itself.

Hectate: OK, I see what you mean there. I'll pull those out.

geekingfrog (approved):

PR #139, opened 2026-05-13
Added battle/ended event

### LINKED ISSUE #48: Add a `battle/ended` event
The event should be sent to all players involved in a battle (including spectators, although we don't currently support them).
It should contain the list of players involved, and who won.

BElluu:
Closed #48

### DIFF
--- a/src/schema/battle/ended.ts
+++ b/src/schema/battle/ended.ts
@@ -1,0 +1,26 @@
+import Type from "typebox";
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "Sent to all players and spectators involved in a battle when it ends.",
+    event: {
+        data: Type.Object({
+            battleId: Type.String({ format: "uuid" }),
+            players: Type.Array(
+                Type.Object({
+                    userId: Type.Ref("userId"),
+                    allyTeamId: Type.String(),
+                })
+            ),
+            spectators: Type.Array(Type.Ref("userId")),
+            winningAllyTeamIds: Type.Optional(
+                Type.Array(Type.String(), {
+                    description: "absent or empty array indicates a draw",
+                })
+            ),
+        }),
+    },
+});
--- a/src/schema/battle/ended.ts
+++ b/src/schema/battle/ended.ts
@@ -12,7 +12,7 @@ export default defineEndpoint({
             players: Type.Array(
                 Type.Object({
                     userId: Type.Ref("userId"),
-                    allyTeamId: Type.String(),
+                    allyTeam: Type.String(),
                     team: Type.String(),
                     player: Type.String(),
                 })
@@ -22,7 +22,7 @@ export default defineEndpoint({
                     userId: Type.Ref("userId"),
                 })
             ),
-            winningAllyTeamIds: Type.Array(Type.Integer()),
+            winningAllyTeamIds: Type.Array(Type.String()),
         }),
     },
 });
--- a/src/schema/battle/ended.ts
+++ b/src/schema/battle/ended.ts
@@ -8,19 +8,21 @@ export default defineEndpoint({
     description: "Sent to all players and spectators involved in a battle when it ends.",
     event: {
         data: Type.Object({
-            battleId: Type.String({ format: "uuid" }),
+            battleId: Type.String(),
             players: Type.Array(
                 Type.Object({
                     userId: Type.Ref("userId"),
                     allyTeamId: Type.String(),
+                    team: Type.String(),
+                    player: Type.String(),
                 })
             ),
-            spectators: Type.Array(Type.Ref("userId")),
-            winningAllyTeamIds: Type.Optional(
-                Type.Array(Type.String(), {
-                    description: "absent or empty array indicates a draw",
+            spectators: Type.Array(
+                Type.Object({
+                    userId: Type.Ref("userId"),
                 })
             ),
+            winningAllyTeamIds: Type.Array(Type.Integer()),
         }),
     },
 });
--- a/docs/schema/autohost.md
+++ b/docs/schema/autohost.md
@@ -91,7 +91,7 @@ Request to add a new player to the battle.
             "title": "AutohostAddPlayerRequestData",
             "type": "object",
             "properties": {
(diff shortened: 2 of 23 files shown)


src/schema/battle/ended.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { defineEndpoint } from "@/generator-helpers.js";
+
+export default defineEndpoint({
+    source: "server",
+    target: "user",
+    description: "Sent to all players and spectators involved in a battle when it ends.",
+    event: {
+        data: Type.Object({
+            battleId: Type.String({ format: "uuid" }),
```
geekingfrog: ditch the format, any string is a valid ID

src/schema/battle/ended.ts
```diff
@@ -0,0 +1,26 @@
...
+            players: Type.Array(
+                Type.Object({
+                    userId: Type.Ref("userId"),
+                    allyTeamId: Type.String(),
```
geekingfrog: There should also be `team` and `player`, these are the 3 parameters identifying a player.

src/schema/battle/ended.ts
```diff
@@ -0,0 +1,26 @@
...
+                })
+            ),
+            spectators: Type.Array(Type.Ref("userId")),
```
geekingfrog: let's make this an array of object, if we want to add more data later, like `joinedAt` or `leftAt`.

src/schema/battle/ended.ts
```diff
@@ -0,0 +1,26 @@
...
+            winningAllyTeamIds: Type.Optional(
+                Type.Array(Type.String(), {
+                    description: "absent or empty array indicates a draw",
```
geekingfrog: where have you seen that draws are possible? Also, there is no semantic difference between null and empty array, so let's not allow both.

BElluu: I haven't seen it. I just thought it might be possible :)

Hectate: Will this event be for when the match is canceled? I see that happen a lot when a force start happens and someone is AFK and then bail during the pre-match. Since it unbalances the team, nobody wants to play. I don’t know if that hits Draw status or if there’s a winner (or a completely different event).

BElluu: Logically, if a player disconnects, for example, in the first two minutes or doesn't connect at all, the team should be able to cancel the match. I don't think anyone wants to play a game with a deficit or advantage. Whether we call it a draw, cancel, or something else seems irrelevant.

src/schema/battle/ended.ts
BElluu: I changed it, but you have to consider whether it is correct. 
For example In /autohost/update.ts is  `battleId: Type.String({ format: "uuid" })`

src/schema/battle/ended.ts
BElluu: Added

src/schema/battle/ended.ts
BElluu: Changed to array obj

src/schema/battle/ended.ts
BElluu: Ok I changed that. If we ever need draw or match cancellation support, we'll expand the structure :)

src/schema/battle/ended.ts
geekingfrog: hum, that's a mistake, the format thing should not be there. We are not taking advantage of the uuid nature of the id at all. I think I remember looking at that in the past but obviously nothing was done.

src/schema/battle/ended.ts
geekingfrog: Ah yeah, that's a good point, ties aren't possible by design, but if the match is stopped then there is indeed no winner.

src/schema/battle/ended.ts
geekingfrog: I think it should be `allyTeam`, to match `team` and `player`. Implicitly they are ids. Elsewhere (in the lobby) we are using these names, though I don't mind slapping `Id` at the end of each one.

src/schema/battle/ended.ts
```diff
@@ -0,0 +1,28 @@
...
+                    team: Type.String(),
+                    player: Type.String(),
+                })
+            ),
+            spectators: Type.Array(
+                Type.Object({
+                    userId: Type.Ref("userId"),
+                })
+            ),
+            winningAllyTeamIds: Type.Array(Type.Integer()),
```
geekingfrog: This is inconsistent with the type of the id as mentioned in `players`. We are using json merge patch for lobby updates, which means array are represented by objects and the keys can be sorted. So that should be an array of string.
I am not sure if this should also be an object so we can add more things to it later though :/

src/schema/battle/ended.ts
BElluu: Okay, so maybe we should organize this to make it consistent? I see that throughout the autohost "module," the battleID is in UUID format. Shall we remove it so that everywhere the battleID is simply sring?

src/schema/battle/ended.ts
BElluu: In the "autohost" module, there is winningAllyTeams, which is an integer. Perhaps we should standardize it to be an object, and TeamId to be a string?

image

geekingfrog: That's because the autohost is a more direct interface with the engine, and the index there correspond to what was sent in `autohost/start` where the team config is passed as an array.
There is indeed a disconnect between the two. However, since the lobby only deal with strings for team ids, I think it makes sense to keep this as well since `battle/ended` is meant for clients. The battle process or whatever in teiserver will have to do the translation.

src/schema/battle/ended.ts
geekingfrog: yeah, we should do that. I had a quick look at the recoil autohost code, and it seems it treats the battle id as string, so removing the uuid format should be fine.

geekingfrog:
CI not happy :(
Otherwise I think the changes are good to go.

BElluu:
Ok I will check it when back to home :)

BElluu:
@geekingfrog re-run CI, should be better :)

Hectate (approved):
Looks good and I see a concur, approved.

PR #141, opened 2026-05-26
Telemetry schemas

### LINKED ISSUE #140: Telemetry
Implementation schemas for Telemetry. Needed for https://github.com/beyond-all-reason/bar-lobby/issues/54

BElluu:
Closed #140 

Adds four new Tachyon endpoints under the telemetry/ namespace to support client-side analytics in bar-lobby:

- `telemetry/simpleClientEvent` - log a named event with no payload
- `telemetry/complexClientEvent` - log a named event with an arbitrary JSON payload (max 4096 bytes)
- `telemetry/updateClientProperty` - set or overwrite a named client property (e.g. window mode, resolution)
- `telemetry/uploadInfolog` - upload a gzip-compressed, base64-encoded log file; responds with infoLogId


src/schema/telemetry/uploadInfolog.ts
```diff
@@ -0,0 +1,35 @@
...
+            }),
+            anonHash: Type.String({
+                description: "Unique install hash (used for both authed and anon users).",
+            }),
+            metadata: Type.Optional(
+                Type.Object({}, { description: "Arbitrary metadata key-value pairs." })
+            ),
+            contents: Type.String({
+                description: "Base64-encoded, gzip-compressed log contents.",
+            }),
```
geekingfrog, line 23: that's not going to fly. Tachyon (and anything json based in general) is absolutely not meant to move files around. This can be big as well, making it really really not a good fit for that kind of thing.

src/schema/telemetry/updateClientProperty.ts
geekingfrog, line 1: what is this about? it says "set or overwrite", how do you retrieve it? I don't get how you do any of that.

src/schema/telemetry/simpleClientEvent.ts
geekingfrog, line 1: I don't think we need a different message between simple and complex. The payload can be optional, and that's it.

geekingfrog (commented):
In general, I don't know where any of these requirements came from.
Like, what things are we trying to know? What features do we want?
The whole thing about `anonHash` doesn't make sense, there is no concept of unauthenticated tachyon connection, at least yet. If we allow that, that would need additional considerations and discussion.

That read like some hallucination without much thought behind.

BElluu:
Overall, it needs to be discussed. I based it solely on the task and simply wanted to approach the discussion with anything, with any vision. The discussion is open ☺️

p2004a:
It does not make sense to spend a second looking at PRs like this without prior design.

I'm not convinced that telemetry should be even going via tachyon at all.

PR #144, opened 2026-06-18
Add unit restrictions to lobbies

geekingfrog:
Similar to game options, but it's represented differently in the start script, so we need something at the proto level as well.

### DIFF
--- a/docs/schema/lobby.md
+++ b/docs/schema/lobby.md
@@ -546,6 +546,11 @@ Create a lobby
                         }
                     }
                 },
+                "restrictions": {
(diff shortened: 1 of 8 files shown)


Hectate:
I think I know how this is supposed to work, but want to make sure:
For each unit restriction the client wants, the key will be the unit name and the value will be a number representing the max number of those units? Set to zero and that unit is unavailable entirely for the match, right?

``armflea: 0`` will disable Armada Tick, for example.

geekingfrog:
I think so? I'm just following the format from the autohost/start schema

Hectate (approved):

PR #146, opened 2026-06-26
Added 'checkAssets' request

### LINKED ISSUE #145: Party members have no way to communicate readiness for queueing
Per docs, if any member of a party sends `matchmaking/queue`, all members will also auto-queue and receive a `matchmaking/queuesJoined` event. A side effect of this is that if they do not have all the assets for the requested queue their game will crash on attempting to launch the battle.

We need some way for a client to communicate what its readiness is in relation to the potential queues, so it can be shared with other party member's clients, and potentially warn/block joins if appropriate.

We have `lobby/updateClientStatus` for lobbies, but unlike a lobby the party has no knowledge of the matchmaking queue(s) currently.

My thought is that we could add to the partyState like so:
```ts
...
    members: {
        userId: UserId;
        joinedAt: UnixTime;
        queueReadiness?: {
            [k:string]:{
                isReady?:boolean,
                assetStatus?: "missing" | "downloading" | "complete";
    }[];
```
The implied purpose of the `[k:string]` is to hold the queue ID for each queue, but the server will not need to validate it. Instead, each client can do a check against it's own list of known queues.

Of course this means we also need to add a `matchmaking/updateClientStatus` request to add to go along with it.

Hectate:
Closes #145 

Allows server to request client status for the list of assets in a specific matchmaking queue prior to adding them to the queue.


geekingfrog:
Since we have specific error codes for `matchmaking/queue`, I think we need an additional one now, like `party_member_missing_asset` or something along those lines to tell the initiator why the request failed.

Hectate:
> Since we have specific error codes for `matchmaking/queue`, I think we need an additional one now, like `party_member_missing_asset` or something along those lines to tell the initiator why the request failed.

Done

geekingfrog (approved):

PR #157, opened 2026-09-17
Let user/report name a battle

burnhamrobertp:
Adds an optional battleId to user/report so a report can say which game it is about, and bounds userIds and message.

Nothing consumes battleId yet. teiserver has no user/report handler on main at all (that is beyond-all-reason/teiserver#1564, still open), so this is the protocol side going first and probably wants to wait for the server side before merging.

battleId rather than a match reference because it is what a client already holds, arriving in privateBattle at battle start and again in battle/ended, so someone reporting a teammate straight after a game can still name it. Optional because a report filed from a lobby has no battle to point at, and the server cannot work one out either - a tachyon lobby has no match until a battle actually starts.

AI disclosure: written with assistance from Claude Code.


L-e-x-o-n:
Battleid is a match reference (battle = match).

Maybe this is a good opportunity to add report subtype as well?

burnhamrobertp:
> Maybe this is a good opportunity to add report subtype as well

My initial new-lobby UX had subType, just like the current web implementation, but moderation has explicitly expressed interest in narrowing down to just the two main types. 🤷

L-e-x-o-n (approved):

Hectate (approved):

